PR #1, opened 2016-07-02
Jeff/update hooks

sofuture:
using new hooks format + resources


r0mant:
Do you really need a dedicated bootstrap container now? As far as I see you just use `kubectl create` so you should be good with just using the default `hook:0.0.1` image. It provides `/usr/local/bin/kubectl`.

sofuture:
Good point, I'll fix.

r0mant:
lgtm

@sofuture Thanks for taking care of this!

alexey-medvedchikov:
Works for me

PR #5, opened 2016-07-20
Make single point where all artifacts and app itself version specifie…

alexey-medvedchikov:
…d: root Makefile

sed used for templating and it's questionable.


resources/app.yaml.tpl
```diff
@@ -0,0 +1,41 @@
+apiVersion: v1
```
klizhentas, line 1: `VERSION` does not have to be template, `app import` has `--version` flag that sets the version for the imported application

sofuture:
Same issue here as: https://github.com/gravitational/monitoring-app/pull/5

This is better solved with real YAML files + Makefile only overrides on import as @klizhentas mentioned.

See here for example: https://github.com/gravitational/pithos-app/blob/master/Makefile

alexey-medvedchikov:
Made #6 instead

PR #6, opened 2016-07-21
Better version control (like in pithos-app)

alexey-medvedchikov:
Gravity 0.0.30 required

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+export VERSION := 0.0.4
+REPOSITORY := gravitational.io
+NAME := logging-app
+OPS_URL ?= https://opscenter.localhost.localdomain:33009
+OUT ?= $(NAME).tar.gz
+
 .PHONY: package
 package:
 	$(MAKE) -C images all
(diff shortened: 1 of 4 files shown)


sofuture:
Do we really want to always build the tag 'latest' instead of a specific version? I could see either way, but curious for your reasons or thoughts @alexey-medvedchikov 

:shipit: in general though

sofuture:
I think I am pro `:latest` now, in conjunction with `--set-image` (doing this in pithos-app right now).

Makefile
```diff
@@ -6,3 +12,11 @@ package:
 deploy:
 	$(MAKE) -C images deploy
 
+.PHONY: import
+import: package
+	-gravity app delete --ops-url=$(OPS_URL) $(REPOSITORY)/$(NAME):$(VERSION) \
+		--force --insecure
+	gravity app import --vendor --glob=**/*.yaml --registry-url=apiserver:5000 \
+		--ops-url=$(OPS_URL) --repository=$(REPOSITORY) --name=$(NAME) \
+		--version=$(VERSION) --insecure .
```
klizhentas, line 22: I would also add `--set-image=log-collector:$(VERSION)` and `--set-image=log-forward:$(VERSION)` to set proper image names during build.

This also means that during container builds you'd need to tag resulting image both as `log-forward:$(VERSION)` - to vendor it and `log-forward:latest` to use in development mode directly.

alexey-medvedchikov:
@sofuture there's more: it's not working that way. Fixed by passing VERSION from root Makefile

sofuture:
shipit

PR #8, opened 2016-07-25
log-tailer component

a-palchikov:
This PR adds a `log-tailer` component that tails logs resulting from the `log-collector`. The new container shares the pod with the `log-collector` and listens on port `8083` by default.

The tailer has a very basic mode of operation: it has a single websocket endpoint `/ws`. Upon client connection, it reads a free-text filter query, builds a tailing pipe using `tail` and, optionally, `grep` and forwards the output of this process group to the client.
The tailer tries to interpret the filter as a structured query before falling back to using it verbatim.

The filter format is as follows:
  `object-type-prefix:name (and object-type-prefix:name)*`

where `object-type-prefix` is one of:
- `pod` to filter only messages from the specific pod
- `container` to filter only messages from the specific container

When multiple containers have been specified, log output from all is forwarded. With a pod filter, only logs of containers from the specified pod are output.


cmd/wstail/matcher.go
```diff
@@ -0,0 +1,43 @@
...
+
+	var podMatchers []string
+	for _, podName := range pods {
+		podMatchers = append(podMatchers, fmt.Sprintf(`%v_%v_%v`, podName, podNamespace, containerMatcher))
+	}
+	return "^" + prefix + match.whitespace + "(" + strings.Join(podMatchers, "|") + ")"
+}
+
+// Macthers
+var match = struct {
```
klizhentas: why not make it a permanent struct?

klizhentas: or simply constants will do too

a-palchikov: i kind of wanted the scope with `match` but I think you're right - scratch that, constants will do

Makefile
```diff
@@ -17,8 +17,12 @@ deploy:
 import: package
 	-$(GRAVITY) app delete --ops-url=$(OPS_URL) $(REPOSITORY)/$(NAME):$(VERSION) \
 		--force --insecure
-	$(GRAVITY) app import --debug --vendor --glob=**/*.yaml --registry-url=apiserver:5000 \
-		--set-image=log-collector:$(VERSION) --set-image=log-forwarder:$(VERSION) \
-		--ops-url=$(OPS_URL) --repository=$(REPOSITORY) --name=$(NAME) \
-		--version=$(VERSION) --insecure .
+	$(GRAVITY) app import --debug --insecure --vendor --glob=**/*.yaml \
+		--registry-url=apiserver:5000 --ops-url=$(OPS_URL) \
+		--set-image=log-collector:$(VERSION) --set-image=log-forwarder:$(VERSION) --set-image=log-tailer:$(VERSION) \
```
sofuture: Minor nitpick: Maybe collect some or all of these args in a variable (especially `--set-image` ones).

cmd/wstail/query.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/gravitational/trace"
+)
+
+// parseQuery interprets the specified reader as a query string and
+// splits it into parts for each search component
+func parseQuery(r io.Reader) (filter filter, err error) {
+	s := bufio.NewScanner(r)
+	s.Split(bufio.ScanLines)
+	d := map[string][]string{}
```
sofuture: More clear name?

cmd/wstail/query.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+type term struct {
+	name  string
+	value string
+}
+
+type parser struct {
+	errors []error
+	s      scanner.Scanner
```
sofuture: IMHO these `parser` member names are too short.

cmd/wstail/query.go
```diff
@@ -0,0 +1,154 @@
...
+	lit    string
+}
+
+type operator string
+
+const (
+	opAND operator = "and"
+)
+
+func (r *parser) next() {
```
sofuture, line 76: `p *parser`?

cmd/wstail/query.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func (r *parser) parseIdent() string {
+	name := r.lit
+	r.expect(scanner.Ident)
+	return name
+}
+
+func (r *parser) parseIndentOrLit() (value string) {
```
sofuture: `Literal`?

cmd/wstail/query.go
```diff
@@ -0,0 +1,154 @@
...
+	switch op {
+	case opAND:
+	default:
+		r.error(r.pos, fmt.Sprintf("expected an operator but got %v", r.lit))
+	}
+	r.expect(scanner.Ident)
+	return op
+}
+
+func (r *parser) expect(tok rune) {
```
sofuture: maybe `token` instead of `tok`?

cmd/wstail/query.go
```diff
@@ -0,0 +1,154 @@
+package main
```
sofuture, line 1: `op` and `pos` are expected abbreviations, but I'd prefer if `s`, `d`, `p`, `tok`, `lit` were all expanded.

cmd/wstail/tail.go
```diff
@@ -0,0 +1,242 @@
...
+	messageC := make(chan []byte)
+	go func() {
+		s := bufio.NewScanner(r)
+		s.Split(bufio.ScanLines)
+		for s.Scan() {
+			line := s.Bytes()
+
+			// Skip to the actual log message in the stream
+			var logEntryPos int
+			for i := 0; i < 3; i++ {
```
sofuture: What's `3` here?

cmd/wstail/tail.go
```diff
@@ -0,0 +1,242 @@
...
+	}()
+	return notifierC
+}
+
+// newMessagePump spawns a goroutine to handle messages from the tailing process group.
+// Returns a channel where the received messages are sent to.
+func newMessagePump(r io.Reader) chan []byte {
+	messageC := make(chan []byte)
+	go func() {
+		s := bufio.NewScanner(r)
```
sofuture, line 118: IMHO this is a fine use of `s` since it's not referenced beyond the next handful of lines, just to frame my comments in other places.

sofuture:
Bunch of style nitpicks in `query.go`, but generally :+1:

cmd/wstail/query.go
a-palchikov, line 76: I usually choose `r` for receiver regardless of the context. I won't make a stand for it but it definitely makes reasoning a bit simpler.

sofuture:
:shipit:

PR #9, opened 2016-07-26
Dmitri/tail initial history

a-palchikov:
This PR adds support for smarter initial history.
The original attempt at providing logging history used `tail -nK` with some arbitrary `K` to capture the last `K` lines of log messages from the aggregated storage (`/var/log/messages`). The problem with this approach is that it was not filter-sensitive and usually did not capture any (enough) information about the filtered data.
An alternative approach is running a `grep` with the specified filter first, tailing the results to some arbitrary `K` (and thus capturing only relevant details), then have the second `tail`/`grep` pipeline take care of the appended messages.

### DIFF
--- a/cmd/wstail/tail.go
+++ b/cmd/wstail/tail.go
@@ -42,9 +42,15 @@ func writer(ws *websocket.Conn, filter filter) {
 	log.Infof("active filter: %v (%v)", filter, matcher)
 	defer ws.Close()
 
-	tailCmd := exec.Command("tail", fmt.Sprintf("-n%v", tailHistory), "-f", filePath)
+	var err error
+	var history []byte
+	tailCmd := exec.Command("tail", "-f", filePath)
 	commands := []*exec.Cmd{tailCmd}
 	if matcher != "" {
+		history, err = snapshot(matcher, filePath, tailHistory)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/wstail/tail.go
```diff
@@ -42,9 +42,15 @@ func writer(ws *websocket.Conn, filter filter) {
...
 
-	tailCmd := exec.Command("tail", fmt.Sprintf("-n%v", tailHistory), "-f", filePath)
+	var err error
+	var history io.ReadCloser
+	tailCmd := exec.Command("tail", "-f", filePath)
 	commands := []*exec.Cmd{tailCmd}
 	if matcher != "" {
+		history, err = snapshot(matcher, filePath, tailHistory)
+		if err != nil {
+			log.Errorf("failed to obtain %v lines of history for %v: %v", tailHistory, matcher, trace.DebugReport(err))
```
sofuture: This is probably generally a warn (i.e. not enough history) and in the case that it's an actual error, should we abort here?

sofuture:
:shipit: aside from a question

cmd/wstail/tail.go
a-palchikov: Showing history is not vital - the tail aborts only if the actual tailing fails. With no history obtained, there's simply no historical output.

PR #10, opened 2016-07-26
Fix a data race in using []byte slices from multiple goroutines:

a-palchikov:
scanner.Bytes() returns a reference to internal reusable buffer which should be copied prior to forwarding it to another goroutine for processing.
Additionally, unwrap JSON-encoded log messages before sending them as logging payload to client.


sofuture:
:+1: lgtm

r0mant:
lgtm

PR #13, opened 2016-07-28
Dmitri/fix process close

a-palchikov:
Explicitly close the process group used to retrieve history after the history has been consumed to prevent the processes (`<defunct>`) from hanging in background.
Also, close all open stdout handles explicitly upon termination.


sofuture:
LGTM

PR #14, opened 2016-07-29
Dmitri/forwarder config

a-palchikov:
This PR adds support for log forwarder configuration to the `log-collector`. In order to achieve this, `log-tailer` has moved inside the `log-collector` container to be able to directly control `rsyslogd`.

The new `/forwarders` endpoint accepts PUT requests with a list of forwarders which it then writes as `rsyslogd` .conf files and restarts the service.


cmd/wstail/forwarders.go
```diff
@@ -0,0 +1,101 @@
...
+			return trace.Wrap(err)
+		}
+		if path == rsyslogConfigDir {
+			return nil
+		}
+		if info.IsDir() {
+			return filepath.SkipDir
+		}
+		log.Infof("removing %v", path)
+		return os.Remove(path)
```
klizhentas: `return trace.Wrap(os.Remove(path), "failed removing %v", path`

cmd/wstail/forwarders.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	// Write new forwarder configuration
+	for _, forwarder := range forwarders {
+		f, err := os.Create(path(forwarder))
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		_, err = f.WriteString(config(forwarder))
+		f.Close()
```
klizhentas: check for error

cmd/wstail/forwarders.go
```diff
@@ -0,0 +1,101 @@
...
+		log.Warningf("failed to delete forwarder configuration files: %v", err)
+	}
+
+	// Write new forwarder configuration
+	for _, forwarder := range forwarders {
+		f, err := os.Create(path(forwarder))
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		_, err = f.WriteString(config(forwarder))
```
klizhentas, line 55: check for error

cmd/wstail/forwarders.go
```diff
@@ -0,0 +1,101 @@
...
+		f.Close()
+		if err != nil {
+			return trace.Wrap(err)
+		}
+	}
+
+	// Reload rsyslogd
+	if out, err := exec.Command("/etc/init.d/rsyslog", "restart").CombinedOutput(); err != nil {
```
klizhentas: what's the base container for this? are we using @alexey-medvedchikov 's containers?

klizhentas: why is this not a constant, while `const rsyslogConfigDir` is?

a-palchikov: the base container is `gravitational/debian-grande`

klizhentas: so you are using `dumb-init` there or something else?

klizhentas: how is reload is going to work if you don't use init.d and using dumb-init?

cmd/wstail/forwarders.go
a-palchikov, line 55: checking down the line

cmd/wstail/forwarders.go
sofuture: I suspect the files are there, but yeah, won't work. Might be able to just `hup` the rsyslog pid though.

klizhentas:
Also, I think you should make API versioned

PR #15, opened 2016-08-02
Dmitri/log link

a-palchikov:
This PR adds support for a side-car container to help transport custom application log files in the format expected by a log-forwarder - as symlinks to actual log file(s) named using the same scheme as used by kubelet for docker logs.
Additionally, support for `file` filter has been added.
Please, see images/loglinker/README.md for details.


sofuture:
Did you run into inotify limits while testing? Even though this sidecar is only going to read specified directories, it's still pretty easy to blow through the default allocation of inotify watchers. May be prudent to sysctl a higher limit when running.

I know that's not super nice and tidy, but while docker exposes some sysctl params (not sure if inotify handles is one of them), k8s does not (yet).

sofuture:
No comments on code or docs beyond inotify limits, so :shipit: after considering those.

klizhentas:
lgtm

a-palchikov:
Frankly - have not seen any issues while testing. Then again, my test cases were pretty rudimentary with the biggest one in clustrix app. But I would be wary of doing anything to counter problems that do not exist just yet.

PR #17, opened 2016-09-02
Bump go version to 1.7 in buildbox and update the go get for godep to

a-palchikov:
explicitly specify binary install directory.


sofuture:
:shipit:

PR #18, opened 2016-09-12
Extended the query syntax to allow `-` and `.` as identifier characte…

a-palchikov:
…rs to simplify pod/container specification.
Added test for the query parser.
Fixed matcher test broken after the addition of the `file` query term.


sofuture:
LGTM :shipit:

PR #20, opened 2016-09-28
Add API method for downloading aggregated logs

r0mant:


cmd/wstail/tail.go
```diff
@@ -273,6 +275,26 @@ func serveWs(w http.ResponseWriter, r *http.Request) (err error) {
 	return nil
 }
 
+// downloadLogs serves /v1/download
+//
+// it compresses the file with aggregated logs and pushes it to the response
+// stream
+func downloadLogs(w http.ResponseWriter, r *http.Request) error {
+	file, err := os.Open(filePath)
```
a-palchikov: Do you want to grab the rotated messages as well or just the active one?

r0mant: I guess we could include older logs too. Would they be named just like `filePath+.<number>` (e.g. `/var/log/messages.1, /var/log/messages.2`) or they are named differently?

a-palchikov (commented):
One question regarding the need to archive rotated messages - otherwise lgtm

cmd/wstail/tail.go
a-palchikov: yeah, I think it's about right - I'm not 100% sure, but it should be standard (`savelog`)

r0mant:
@a-palchikov I have updated the PR to collect all logs. You might wanna take another look cause the implementation changed significantly :)

r0mant:
Anyway, I've tested it and it works cool so I'm gonna merge.

PR #22, opened 2016-10-13
Bump remote_syslog to v0.18+ 

a-palchikov:
- (a few commits after v0.18) to include a change when remote_syslog2 forwards new files entirely (as opposed to tailing them). This is necessary for static logs as these need to be forwards in their entirety.
- Bumping above also discovered the [issue](https://github.com/golang/go/issues/16739) with go1.7 net module ignoring `.local` suffixed domain names from DNS lookup and updated to go1.7.1 where this has been fixed.
- Tweaked matcher w.r.t to files to also handle files outside of k8s context - as the case is with external logs not from a container
- Removed querying the log history - simply using `tail -n +1` will start tailing from the beginning
- `tail` will also output rotated log files (if any) prior to tailing the active one

### DIFF
--- a/cmd/wstail/tail.go
+++ b/cmd/wstail/tail.go
@@ -75,7 +75,7 @@ func tailer(ws *websocket.Conn, filter filter) {
 					if truncAt > maxDumpLen {
 						truncAt = maxDumpLen
 					}
-					log.Infof("failed to unmarshal `%v`: %v", message[:truncAt], err)
+					log.Infof("failed to unmarshal `%v...`: %v", message[:truncAt], err)
 					// Use the message as-is
 				}
 			}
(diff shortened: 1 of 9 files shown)


klizhentas (approved):

sofuture (approved):

cmd/wstail/tail.go
```diff
@@ -24,23 +25,24 @@ import (
...
 	matcher := buildMatcher(filter)
 	log.Infof("active filter: %v (%v)", filter, matcher)
 	defer ws.Close()
 
-	var err error
-	var history io.ReadCloser
-	tailCmd := exec.Command("tail", "-f", filePath)
+	rotated, err := collectRotatedMessages()
+	if err != nil {
+		log.Errorf("failed to read a list of rotated log files: %v", err)
```
klizhentas, line 39: is it ok that you continue here?

a-palchikov: I'll bail out if that happens - this is an error after all.

PR #23, opened 2016-10-14
Tune the logging behavior / switch to a forked remote_syslog (for the time being)

a-palchikov:
- `tail`: do not send the whole aggregated log file on an empty filter query - limit this to some arbitrary number of last lines (100 as used previously for history).
- `forwarder`: switch to a fork of remote_syslog with a fixed problem of initial failure to connect leading to forwarder being unable to subsequently reconnect. This also removes the need for the `livenessProbe`.

Fixes https://github.com/gravitational/gravity/issues/1111 and https://github.com/gravitational/gravity/issues/1119

This PR requires that this be merged as well: https://github.com/papertrail/remote_syslog2/pull/175

A small caveat: if the forwarder starts (i.e. initially successfully connects) **after** any log files have been already placed (i.e. install log -> right after the installation and the forwarder is flaking), they will not be forwarded as new (i.e. in their entirety) - a more complete solution (to remote_syslog2) would be to always start and try to rebuild the top-level connection.
I think I want to implement this and see how that works.

### DIFF
--- a/cmd/wstail/query.go
+++ b/cmd/wstail/query.go
@@ -146,6 +146,13 @@ func (r *parser) error(pos scanner.Position, message string) {
 	r.errors = append(r.errors, trace.Errorf("%v: %v", pos, message))
 }
 
+func (r filter) isEmpty() bool {
+	return len(r.containers) == 0 &&
+		len(r.pods) == 0 &&
+		len(r.files) == 0 &&
+		r.freeText == ""
+}
+
 func tokenStrings(tokens []rune) string {
 	var output bytes.Buffer
 	for i, token := range tokens {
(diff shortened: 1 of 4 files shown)


sofuture (approved):

cmd/wstail/query.go
```diff
@@ -146,6 +146,13 @@ func (r *parser) error(pos scanner.Position, message string) {
 	r.errors = append(r.errors, trace.Errorf("%v: %v", pos, message))
 }
 
+func (r filter) isEmpty() bool {
+	return len(r.containers) == 0 &&
+		len(r.pods) == 0 &&
+		len(r.files) == 0 &&
+		r.freeText == ""
```
r0mant, line 153: How about `strings.TrimSpace(r.freeText)`? How sure if that's important or not in this context.

r0mant:
lgtm

PR #26, opened 2016-11-29
WIP: add update resources using rig

klizhentas:
Please do not merge yet, as I'm still working on rollback.

This PR adds update hook using rig for changeset management


images/hook/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+set -e
+
+if [ $1 = "update" ]; then
+    echo "Starting update, changeset: $RIG_CHANGESET"
+    echo "Deleting old replication controller rc/log-forwarder"
+    rig delete rc/log-collector --resource-namespace=kube-system
+    echo "Creating or updating resources"
+    rig upsert -f /var/lib/gravity/resources/resources.yaml --debug
+    echo "Checking status"
+    rig status $RIG_CHANGESET --retry-attempts=120 --retry-period=1s --debug
```
a-palchikov, line 14: `timeout 120 watch -n 1 -d rig status $RIG_CHANGESET --debug`

resources/app.yaml
```diff
@@ -22,7 +22,25 @@ hooks:
...
+  update:
+    spec:
+      apiVersion: batch/v1
+      kind: Job
+      metadata:
+        name: logging-app-update
+      spec:
+        parallelism: 1
+        completions: 1
+        activeDeadlineSeconds: 100
```
a-palchikov, line 34: if the job takes longer than that, `rig freeze` will not get called, i guess.
  * Without knowing what `rig freeze` does (committing a transaction?), how do we know update was successful?   
  * Does `rig status` terminate early upon success?
  * Does it fail at all or simply dumps information to stdout?
  * If the update fails, will `rig freeze` still be called? What is its behavior in either case (success/failure)?

klizhentas: update is successfull if script does not return error at any stage
freeze simply marks changeset as completed, with no further modifications allowed. 
these are not transactions, simply tracking changesets, changes are applied after each command
rig status will terminate on early success
it fails and dumps info to stdout
rig freeze will be only called on successfull update

images/hook/entrypoint.sh
klizhentas: not sure where are you going with this

a-palchikov: I meant it as a weak attempt to show that `--retry-xyz` were not necessary.

images/hook/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "Freezing"
+    rig freeze
+else
```
sofuture: does this work as intended? it reads weird -- is it supposed to happen on subsequent run after failure of first run?

resources/app.yaml
a-palchikov: This still looks a bit magic (inconsistent) - why does only `rig status` take the changeset as parameter? Is it implicit in all other commands (or not necessary)?

images/hook/entrypoint.sh
klizhentas: I think you missed `set -e` in the begining of the script. In case of failure this line won't be reached.

resources/app.yaml
klizhentas: it takes the value from environment variable for all commands except this one.

images/hook/entrypoint.sh
a-palchikov: It does read weird: invoking the hook w/o arguments will rollback. I guess this is only useful with the rollback hook, since the rollback is not necessary during update in case of a failure (correct me if I'm wrong). If the rollback is to only be used in a rollback hook, then an explicit command would be better than the `else` case.

images/hook/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/dumb-init /bin/sh
```
sofuture: Should just be `/bin/sh`

images/hook/entrypoint.sh
sofuture: I saw the `set -e` which is why I don't understand what this block does, or when it runs. It will call `rig rollback` if the script is run without args? Shouldn't it roll back in case of failure?

images/hook/entrypoint.sh
klizhentas: we need dumb-init to handle signals properly:

https://github.com/Yelp/dumb-init#why-you-need-an-init-system

images/hook/entrypoint.sh
klizhentas: @sofuture I don't understand your question. Can you elaborate?

klizhentas: I guess you are confused because there's no `rollback` hook added yet. I will add it today

sofuture: Okay, that makes a bit more sense. What's responsible for calling rollback hook? Automatically called by platform if update hook fails?

images/hook/entrypoint.sh
sofuture: Agree, but I think it should be only the entrypoint/cmd, and not this file's shebang interpreter.

resources/app.yaml
klizhentas: I added more comments to clarify what's going on

images/hook/entrypoint.sh
klizhentas: what's the difference?

sofuture: It's just unnecessary mixing of concerns IMHO. And I think in this specific case the actual resulting process tree is: `dumb-init` -> `dumb-init` -> `sh`.

klizhentas: good catch, removed extra

Makefile
```diff
@@ -17,6 +17,10 @@ package:
 forwarder:
 	$(MAKE) -C images forwarder
 
+.PHONY: hook
```
sofuture, line 20: Do we actually need any of these image targets split at this level? or can we just have a single `make images` target to call into `images/Makefile`.

Not important either way.

sofuture (approved):
lookin good

Makefile
klizhentas: just for testing/dev

PR #31, opened 2017-01-24
Add hard limit for log filesize

agalitsyn:
add 1.5mb limit for log file, instead read whole file

### DIFF
--- a/cmd/wstail/tail.go
+++ b/cmd/wstail/tail.go
@@ -35,6 +35,14 @@ const tailMaxDepth = 100
 // rotated log file as named by savelog
 const rotatedLogUncompressed = "messages.0"
 
+const (
+	Byte     = 1.0
+	Kilobyte = 1024 * Byte
+	Megabyte = 1024 * Kilobyte
+)
+
+const tailFileSizeLimit = 1.5 * Megabyte
+
 func tailer(ws *websocket.Conn, filter filter) {
 	matcher := buildMatcher(filter)
 	log.Infof("active filter: %v (%v)", filter, matcher)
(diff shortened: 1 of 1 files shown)


cmd/wstail/tail.go
```diff
@@ -35,6 +35,14 @@ const tailMaxDepth = 100
 // rotated log file as named by savelog
 const rotatedLogUncompressed = "messages.0"
 
+const (
+	Byte     = 1.0
```
a-palchikov: Do you need floats just to be able to compute 1.5Mb?

cmd/wstail/tail.go
```diff
@@ -46,18 +54,28 @@ func tailer(ws *websocket.Conn, filter filter) {
...
-	if filter.isEmpty() {
-		// Limit the output of an empty filter to last tailMaxDepth lines
-		tailingDepth = []string{"-n", fmt.Sprintf("%v", tailMaxDepth)}
-	}
 	var files []string
 	if rotated.Main != "" {
 		files = append(files, rotated.Main)
 	}
-	files = append(files, "-F", filePath)
+	files = append(files, "--follow", filePath)
```
a-palchikov: `-F` is not the same as `--follow`

cmd/wstail/tail.go
```diff
@@ -46,18 +54,28 @@ func tailer(ws *websocket.Conn, filter filter) {
...
+
+	f, err := os.Stat(filePath)
+	if err != nil {
+		log.Errorf("cannot get file info for %v: %v", filePath, err)
+		return
```
a-palchikov, line 64: Why exit if the file is not (yet) there?

cmd/wstail/tail.go
```diff
@@ -46,18 +54,28 @@ func tailer(ws *websocket.Conn, filter filter) {
...
+	}
+	size := float32(f.Size())
+	tailingDepth := []string{"--lines", "+1"}
+	var isTrimmed bool
+	if filter.isEmpty() || size > tailFileSizeLimit {
```
a-palchikov: Cannot really use initial log file size as a restriction - the size restriction should ideally be applied to the results after the filter has been applied.

cmd/wstail/tail.go
UnderGreen: @a-palchikov with `--follow` you don't need to use man or help for command to understand what it means.

agalitsyn: true,
```
man tail

-F     same as --follow=name --retry
```

I always try to use long names for options in code for readability.

cmd/wstail/tail.go
agalitsyn: thx for the highlighting this, I I did not know that it should work such way

cmd/wstail/tail.go
agalitsyn: not really, set of consts is easy to read

cmd/wstail/tail.go
agalitsyn: yes, but how achieve this result? pipe goes to scanner, which read it line by line

agalitsyn:
I've fixed review comments

sofuture (approved):
LGTM

a-palchikov:
Querying file size is useless for enforcing this limit.
Besides, size limit should have been applied to the (history + grep) output - not the tail. Limiting tail works w/o regard to the filtered contents and results in outputs much shorter than the advertised limit (100 lines)

agalitsyn:
Thx for clarifying how it exactly should work, I will change behaviour

agalitsyn:
I will open another PR since it's already merged, this stuff will be reverted

PR #32, opened 2017-01-30
Add hard limit for log filesize

agalitsyn:
Partly reverted #31. Now output is limited on subprocess side, using grep & tail subshell


agalitsyn:
closed via https://github.com/gravitational/logging-app/pull/34

PR #33, opened 2017-01-30
Remove unable to parse logs

agalitsyn:
mute this log to prevent garbage in monitoring system

PR #34, opened 2017-01-31
websocket => http

agalitsyn:
Extends both #32 and #33


cmd/wstail/main.go
```diff
@@ -1,34 +1,60 @@
...
 		filePath = defaultTailSource
 	} else {
 		filePath = flag.Args()[0]
 	}
 
-	http.Handle("/v1/ws", makeHandler(serveWs))
-	http.Handle("/v1/download", makeHandler(downloadLogs))
+	log.Infof("HTTP service listening on %s", *httpAddr)
+
+	http.Handle("/v1/log", withContext(filePathContextKey, filePath, makeHandler(getLogs)))
```
a-palchikov: Instead of putting data into context, consider passing `filePath` to a new handler factory function (e.g.`handlerWithFilePath`) to make it explicit - this way you would not need type casts.

cmd/wstail/main.go
```diff
@@ -1,34 +1,60 @@
...
-)
+	signalChan := make(chan os.Signal, 1)
+	signal.Ignore()
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+
+	for {
+		select {
+		case err := <-errChan:
+			if err != nil {
+				log.Fatal(err)
```
a-palchikov: `trace.DebugReport(err)`

cmd/wstail/main.go
```diff
@@ -1,34 +1,60 @@
...
+			}
+		case s := <-signalChan:
+			log.Infof("Captured %v. Exiting...", s)
+			os.Exit(0)
```
a-palchikov: Why explicit `Exit` call? You can simply `return`

cmd/wstail/main.go
```diff
@@ -41,3 +67,11 @@ func makeHandler(handler handlerFunc) http.HandlerFunc {
 }
 
 type handlerFunc func(w http.ResponseWriter, r *http.Request) error
+
+// withContext is a middleware which allows to put data to request context
+func withContext(key, value interface{}, next http.Handler) http.HandlerFunc {
```
a-palchikov: See above - I don't think this is necessary.

cmd/wstail/tail.go
```diff
@@ -13,135 +13,112 @@ import (
...
-	log.Infof("active filter: %v (%v)", filter, matcher)
-	defer ws.Close()
-
+func logReader(filePath string, filter filter, limit string) ([]byte, error) {
 	dir := filepath.Dir(defaultTailSource)
 	names, err := readDir(dir)
 	if err != nil {
-		log.Errorf("failed to read log directory: %v", err)
-		return
+		return nil, trace.Wrap(err, "failed to read log directory: %v")
```
a-palchikov: missing placeholder value in format string

cmd/wstail/tail.go
```diff
@@ -13,135 +13,112 @@ import (
...
+	var commands []*exec.Cmd
+	if filter.isEmpty() {
+		commands = []*exec.Cmd{
+			exec.Command("tail", "--lines", limit, filePath),
 		}
-		history, err = snapshot(matcher, rotated, historyLimit)
+	} else {
+		matcher := buildMatcher(filter)
+		log.Infof("active filter: %v (%v)", filter, matcher)
+		history, err = snapshot(matcher, rotated, -1)
```
a-palchikov: historyLimit was readable, `-1` not so much
Why specify `-1` in the first place - looks like it's a const and can be removed

cmd/wstail/tail.go
```diff
@@ -13,135 +13,112 @@ import (
...
+func logReader(filePath string, filter filter, limit string) ([]byte, error) {
 	dir := filepath.Dir(defaultTailSource)
 	names, err := readDir(dir)
 	if err != nil {
-		log.Errorf("failed to read log directory: %v", err)
-		return
+		return nil, trace.Wrap(err, "failed to read log directory: %v")
+	}
+
+	_, err = os.Stat(filePath)
```
a-palchikov: Generally, I think that failing when reading is more robust as the file might not available when the check is performed and would appear when it's read from.
And the `os.Stat` error message will not descriptive - if the intention is to provide a user-friendly error message for non-existing file, I'd handle the non-existing case with a user-friendly error message.

cmd/wstail/tail.go
```diff
@@ -300,67 +260,29 @@ func (r *processGroup) String() string {
...
-func serveWs(w http.ResponseWriter, r *http.Request) (err error) {
-	ws, err := upgrader.Upgrade(w, r, nil)
-	if err != nil {
-		return trace.Wrap(err, "failed to upgrade to websocket protocol")
-	}
-
-	_, data, err := ws.ReadMessage()
+// getLogs serves /v1/log?query=hello&limit=100
+//
+// it dumps all logs filtered by query and cutted to limit in the configured filePath
```
a-palchikov: s/cutted/cut
`getLogs executes specified query on the log file(s) configured with filePath and returns filtered output optionally truncated to limit. With no limit set, defaults to tailMaxDepth as a limit`

cmd/wstail/tail.go
```diff
@@ -151,30 +128,9 @@ type dockerLogMessage struct {
...
-		close(notifierC)
-	}()
-	return notifierC
-}
-
-// newMessagePump spawns a goroutine to handle messages from the tailing process group.
+// makeOutputChannel spawns a goroutine to handle messages from the process group.
 // Returns a channel where the received messages are sent to.
-func newMessagePump(r io.Reader, history io.ReadCloser) chan string {
+func makeOutputChannel(r io.Reader, history io.ReadCloser) chan string {
```
a-palchikov: Not sure I agree with the name change - it reads as if it only created and returned a channel.

cmd/wstail/tail.go
```diff
@@ -185,7 +141,7 @@ func newMessagePump(r io.Reader, history io.ReadCloser) chan string {
 	// columns to only output the relevant detail
 	const columnsToSkip = 3
 
-	messageC := make(chan string)
+	ch := make(chan string)
```
a-palchikov: same here

cmd/wstail/tail.go
```diff
@@ -288,7 +249,6 @@ func (r *processGroup) Close() (err error) {
 	for _, closer := range r.closers {
 		closer.Close()
 	}
-	r.terminate()
```
a-palchikov, line 291: Why is this not necessary?

agalitsyn: We will have error `wait: no child processes` because process exited cleanly before we call wait.

cmd/wstail/tail.go
agalitsyn: I thought that `trace.ConvertSystemError(err)` will make descriptive error message

cmd/wstail/main.go
agalitsyn: Sure, it's possible, but less flexible. I used context as general-purpose container with versatile middleware, you can set anything later without rewriting function signatures

cmd/wstail/tail.go
agalitsyn:
```
package main

import (
	"fmt"
	"os"

	"github.com/gravitational/trace"
)

func main() {
	filePath := os.Args[1]

	_, err := os.Stat(filePath)
	if err != nil {
		fmt.Println(trace.ConvertSystemError(err))
	}
}
```

`go run main.go /foo`
```
stat /foo: no such file or directory
```

looks ok for me

agalitsyn:
@a-palchikov please check new version, I've fixed your comments

a-palchikov (approved):

cmd/wstail/tail.go
a-palchikov: What if it does not exit fast enough - e.g. if refreshing logs faster than usual - won't this leave (some) as a zombies? It is easier to ignore this error rather than accumulate them, have you tried to repro this?

PR #42, opened 2017-07-10
Mount `/var/log/` in forwarders

a-palchikov:
Loosen the mounts to /var/log to avoid issues with symlinks in k8s 1.7: log files in `/var/log/containers` now link to `/var/log/pods` and the forwarder is not able to read/forward them.

Updates https://github.com/gravitational/gravity/issues/2460

### DIFF
--- a/resources/resources.yaml
+++ b/resources/resources.yaml
@@ -97,21 +97,16 @@ spec:
         - name: log-forwarder
           image: log-forwarder:latest
           volumeMounts:
-            - name: gravitylog
-              mountPath: /var/log/gravity
-            - name: varlogcontainers
-              mountPath: /var/log/containers
+            - name: varlog
+              mountPath: /var/log
             - name: extdockercontainers
               mountPath: /ext/docker/containers
             - name: gravitysite
               mountPath: /var/lib/gravity/site
       volumes:
-        - name: gravitylog
+        - name: varlog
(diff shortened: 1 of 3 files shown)


resources/resources.yaml
```diff
@@ -97,21 +97,16 @@ spec:
...
               mountPath: /var/lib/gravity/site
       volumes:
-        - name: gravitylog
-          hostPath:
-            path: /var/log/gravity
-        - name: varlogcontainers
+        - name: varlog
           hostPath:
-            path: /var/log/containers
+            path: /var/log
```
r0mant, line 109: I remember I was having an issue in monitoring app when I mounted the whole `/etc` into influxdb container (its config was in `/etc/influxdb.conf`). The issue was that during update the old pod couldn't exit because for some reason it apparently tried to wipe the whole /etc during pod teardown (no idea why) and getting "/etc/resolv.conf is busy" or somtething like this. This stopped when I mounted its config to some dir under /etc. Not sure it applies here but just wanted to mention it. Issues may arise later when upgrading from _this_ version.

a-palchikov: I don't think there's any harm in exposing entire `/var/log` (as there was in your case) - although Pod trying to remove the contents of the mounted directory is suspicious.

PR #51, opened 2018-02-09
Remove 32 character default length limit for syslog tags.

a-palchikov:
- Add a verbose flag and make it overridable via a DEBUG environment variable.
 - Use expressive configuration format to be able to specify the formatting template to use for forwarder.

Fixes https://github.com/gravitational/gravity/issues/3106.


a-palchikov:
Wrong branch

PR #52, opened 2018-02-09
Remove 32 character default length limit for syslog tags.

a-palchikov:
- Add a verbose flag and make it overridable via a DEBUG environment variable.
 - Use expressive configuration format to be able to specify the formatting template to use for forwarder.

Fixes https://github.com/gravitational/gravity/issues/3106.

### DIFF
--- a/cmd/wstail/forwarders.go
+++ b/cmd/wstail/forwarders.go
@@ -1,9 +1,11 @@
 package main
 
 import (
-	"fmt"
+	"bytes"
 	"io/ioutil"
+	"net"
 	"path/filepath"
+	"text/template"
 
 	"github.com/ghodss/yaml"
 	"github.com/gravitational/trace"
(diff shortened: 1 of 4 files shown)


cmd/wstail/forwarders.go
```diff
@@ -75,11 +82,30 @@ func forwarderFilename(forwarder logForwarder) string {
 }
 
 // forwarderConfig returns log forwarder rsyslog config
-func forwarderConfig(forwarder logForwarder) []byte {
-	if forwarder.Spec.Protocol == "udp" {
-		return []byte(fmt.Sprintf("*.* @%v", forwarder.Spec.Address))
+func forwarderConfig(forwarder logForwarder) ([]byte, error) {
+	host, port := splitHostPort(forwarder.Spec.Address)
+	var buf bytes.Buffer
+	var config = struct {
```
r0mant: Why not pass a map instead of defining a struct?

```go
err := configTemplate.Execute(&buf, map[string]string{
  "Target": host,
  "Port": port,
  "Protocol": forwarder.Spec.Protocol,
})
```

r0mant (approved):

PR #56, opened 2018-09-27
Set resource quotas.

r0mant:
Set resource requests and limits on forwarders/collectors according to the data provided by a customer in this ticket: https://github.com/gravitational/gravity.e/issues/3436.

### DIFF
--- a/resources/resources.yaml
+++ b/resources/resources.yaml
@@ -87,10 +87,10 @@ spec:
               mountPath: /etc/rsyslog.d
           resources:
             requests:
-              memory: 100Mi
+              memory: 200Mi
               cpu: 100m
             limits:
-              memory: 200Mi
+              memory: 600Mi
               cpu: 500m
       volumes:
         - name: logforwarders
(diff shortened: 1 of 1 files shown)


resources/resources.yaml
```diff
@@ -85,6 +85,13 @@ spec:
               mountPath: /var/log
             - name: logforwarders
               mountPath: /etc/rsyslog.d
+          resources:
+            requests:
+              memory: 100Mi
+              cpu: 100m
+            limits:
+              memory: 200Mi
```
a-palchikov: Just curious - why would you give a collector less memory than the forwarder and in such a proportion?
Also, I know forwarders had been memory-hungrier, maybe they deserve twice as much until we have sorted them out?

a-palchikov (approved):

resources/resources.yaml
r0mant: In the ticket that I referenced the customer mentions that forwarders are memory hungry so I gave them more. Sure, I'll bump the limits for both.

PR #57, opened 2019-05-06
Logrange

kbabushkin:
Introduction of Gravity logging-app based on Logrange streaming database...


cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/jrivets/log4g"
```
r0mant: FYI we're using https://github.com/gravitational/logrus for logging. Could you please refactor the code to use it here as well, for consistency?

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	"github.com/mohae/deepcopy"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+type (
+	Adapter struct {
```
r0mant, line 47: FYI, we're adhering to https://github.com/golang/go/wiki/CodeReviewComments as code style guidelines per which all exported types/methods/vars should have a Godoc (https://github.com/golang/go/wiki/CodeReviewComments#doc-comments).

I'm noticing a lot of exported things do not have Godocs atm.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+		cfg   *Config
+		wTmpl *forwarder.WorkerConfig
+
+		lCli api.Client
```
r0mant: I would use more verbose but clear names for variables, otherwise it's not clear what `lCli` or `kCli` are.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+		r *http.Request, p httprouter.Params) error
+
+	grLogEntry struct {
+		Type    string `json:"type"`
+		Payload string `json:"payload"`
+	}
+)
+
+const (
+	lgLimitLines     = 1000
```
r0mant: These constants here would also benefit from short godocs explaining what they are for.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	dlLimitPerFileBt = 10 * 1024 * 1024
+	dlFilename       = "messages"
+)
+
+func Run(ctx context.Context, cfg *Config, cl api.Client) error {
+	logger := log4g.GetLogger("adapter")
+
+	adaptr, err := NewAdapter(cfg, cl)
+	if err != nil {
+		return fmt.Errorf("failed to create adapter, err=%v", err)
```
r0mant: FYI we're using https://github.com/gravitational/trace to wrap returned errors. You can see examples in our current logging app code, e.g. https://github.com/gravitational/logging-app/blob/master/cmd/wstail/tail.go#L92. The gist is that you wanna do `return trace.Wrap(err)` if there's an error available (which makes it easy to print stacktrace afterwards), or return an appropriate error from trace package, e.g. `trace.BadParameter("")` (most often, used as a generic error), `trace.NotFound("")`, etc.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+func (ad *Adapter) Close() error {
+	var err error
+	if !utils.WaitWaitGroup(&ad.waitWg, time.Minute) {
+		err = errors.New("close timeout")
+	}
+	ad.logger.Info("Closed, err=", err)
+	return nil
+}
+
+//===================== adapter.api =====================
```
r0mant: Nit: I think that divider-comments like this in a source file are a pretty good sign that separated things deserve to be in separate modules (files).

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+	logger.Info("Shutdown.")
+	return err
+}
+
+//===================== adapter =====================
+
+func NewAdapter(cfg *Config, cli api.Client) (*Adapter, error) {
+	f := new(Adapter)
+	f.cfg = deepcopy.Copy(cfg).(*Config)
```
r0mant: Why do you need deepcopy here? Is the config expected to be modified after it has been passed here? If so, I think this is a pretty unexpected side effect.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	var err error
+	ad.kCli, err = newKCli(ad.cfg)
+	if err != nil {
+		return err
+	}
+	ad.wTmpl, err = ad.cfg.Logrange.GetForwarderTmpl()
+	if err != nil {
+		return err
+	}
+	return err
```
r0mant: Nit: `return nil`.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	qLimit := strings.TrimSpace(rq.URL.Query().Get("limit"))
+
+	var (
+		err error
+	)
+
+	limit := lgLimitLines
+	if qLimit != "" {
+		limit, err = strconv.Atoi(qLimit)
+		if err != nil || limit < 0 {
```
r0mant: I would at least log the error.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	qr := &api.QueryRequest{
+		Query: buildLql(qQuery, ad.cfg.Logrange.Partition, limit, -limit),
+		Pos:   "tail", Offset: -limit, Limit: limit,
+	}
+
+	ad.logger.Info("log(): Query=", qr.Query)
+	rs := &api.QueryResult{}
+	err = ad.lCli.Query(ctx, qr, rs)
+	if err == nil {
+		grLogEntries, err := ad.toGrLogEntries(rs.Events)
```
r0mant: This error is ignored. (And it won't be logged below because it overshadows err from parent context). Same for below.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+		ad.waitWg.Add(1)
+		go func(cq cronQuery) {
+			defer ad.waitWg.Done()
+			ad.runCronQuery(ctx, cq)
+		}(cq)
+	}
+}
+
+func (ad *Adapter) runCronQuery(ctx context.Context, cq cronQuery) {
+	ad.logger.Info("Entering into the loop to run \"", cq.Query, "\" every ", cq.PeriodSec, " seconds.")
```
r0mant: Nit: You can use string formatting so you won't have to escape quotes:

```go
xxx.Infof("Entering into the loop top run %q every %q seconds.", cq.Query. cq.PeriodSec)
```

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	for {
+		sleepDelay := time.Second * time.Duration(cq.PeriodSec)
+		res, err := ad.lCli.Execute(ctx, api.ExecRequest{Query: cq.Query})
+		if err != nil {
+			ad.logger.Warn("Could not connect to the server to run ", cq)
+			sleepDelay = time.Second
+		}
+
+		if res.Err != nil {
```
r0mant, line 170: Does `ad.lCli.Execute` return non-nil "res" even if the error occurred? If no, then this will panic.

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+		Namespace              string
+		ForwarderConfigMapName string
+	}
+
+	// cronQuery describe a query, which should be called periodically
+	cronQuery struct {
+		// Query contains LQL which should be called periodically
+		Query string
+		// PeriodSec defines an interval in seconds to execute the Query
+		PeriodSec int
```
r0mant, line 45: Why is this not a `time.Duration`? You wouldn't have to convert it then (I saw it in a few places).

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+}
+
+func (c *Config) Apply(other *Config) {
+	if other == nil {
+		return
+	}
+
+	if other.Gravity != nil {
+		c.Gravity.apply(other.Gravity)
```
r0mant: Do you need to check that `c.Gravity` is also not nil? Same for below.

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+func (l *logrange) apply(other *logrange) {
+	if other == nil {
+		return
+	}
+
+	if other.Transport != nil {
+		l.Transport.Apply(other.Transport)
+	}
+	if other.Kubernetes != nil {
+		l.Kubernetes.apply(other.Kubernetes)
```
r0mant: Looking at these config structs and their methods, I feel like with so many config fields being pointers and the lack of nil checks, something somewhere will definitely panic :) Could you please make sure all nils are handled properly and either add explicit nil checks (so you don't have to look into the actual method to see if it handles nil properly) or add tests for these cases?

resources/app/log-collector.yaml
```diff
@@ -0,0 +1,148 @@
...
+        ],
+        "Kubernetes": {
+          "Namespace": "kube-system",
+          "ForwarderConfigMap": "lr-forwarder"
+        },
+        "Transport": {
+          "ListenAddr": "lr-aggregator.kube-system.svc.cluster.local:9966",
+          "TlsEnabled": false,
+          "Tls2Way": false,
+          "TlsCertFile": "/opt/logrange/gravity/config/tls/client.crt",
```
r0mant: Where do these certificate/key/ca come from? Same q for all other deployments.

resources/app/lr-aggregator.yaml
```diff
@@ -0,0 +1,124 @@
...
+      volumes:
+        - name: config
+          configMap:
+            name: lr-aggregator
+        - name: extdockercontainers
+          hostPath:
+            path: /ext/docker/containers
+        - name: varlib
+          hostPath:
+            path: /var/lib
```
r0mant: Do you need to mount the whole `/var/lib`? Is that where logrange database is stored, `/var/lib/logrange`? If so, I would only mount that.

Same for all other deployments.

README.md
```diff
@@ -1,87 +1,53 @@
...
+# Gravity Log Application
 
-## Overview
+Gravity Log Application is intended for gathering logs and providing API for logs retrieval in Gravity cluster. The application is based on [Logrange](https://github.com/logrange/logrange) streaming database.
 
-There are 2 main components in the logging system: collectors and forwarders.
+### 1. HTTP API
 
-### Forwarder
+Listens *:8083
```
a-palchikov: The server listens on port 8083 by default.

README.md
```diff
@@ -1,87 +1,53 @@
...
 
-The forwarder's role is to read files on disk on each k8s node and forward them
-to a central place. We use [remote_syslog2](https://github.com/papertrail/remote_syslog2)
-to accomplish this, with the following [config](images/forwarder/remote_syslog.yml).
-The forwarder is implemented as a `DaemonSet`.
+#### Query logs
```
a-palchikov: Querying Logs

README.md
```diff
@@ -1,87 +1,53 @@
...
 
-The gist of the forwarders config is to tail all files that from `/var/log/gravity`
-and `/var/log/containers`. Both these paths are mounted from the host the forwarder
-is running on. Kubernetes logs are automatically included via `/var/log/containers`.
+###### GET: /v1/log?limit=&query=
 
-To have your applications logs forwarded from file, you must mount and log to
-`/var/log/gravity`.
+- `query` param terms: "pod:<pod_name>", "file:<container_id>", "container:<container_name>", "and" or "or"
```
a-palchikov: `query` can contain the following terms:
  * `pod:<name>` - to limit search to a specific pod
  * `container:<name>` - to limit search to a specific container inside a pod
  * `file:<file>` - to limit search to a specific log file
  * boolean operators `and` and `or`

README.md
```diff
@@ -1,87 +1,53 @@
...
+- default `limit` is 1000
 
-The collector's role is to collect all incoming logs via an rsyslog server accepting
-both TCP and UDP. The forwarder automatically sends all logs to the service provided
-by the collector.
+- output format: JSON, e.g.: `[{"type":"data","payload":""}]`
 
-To have your application logs directly sent to rsyslog, you must log via rsyslog
-TCP/UDP to `log-collector.kube-system.svc.cluster.local`.
+#### Download (the latest) logs
```
a-palchikov: Downloading ~~(the latest)~~ Logs

README.md
```diff
@@ -1,87 +1,53 @@
...
 
-Logs are currently written to `/var/log/messages` on the host node.
+###### GET: /v1/download
 
-Another facet of the collector container is a simple logging configuration and tailing service.
-The service exposes the following endpoints:
-  - /ws         - websocket endpoint that streams logs to the client
-  - /forwarders - HTTP endpoint for managing log forwarders
+- output format: TEXT files, compressed as `tar.gz` archive
```
a-palchikov: Is it saying that two formats are supported or just one?
```
Streams a compressed tarball.
```

README.md
```diff
@@ -1,87 +1,53 @@
...
 
-The websocket streaming endpoint has a very simple protocol: the first message from the client upon
-connection is a free-form filter query to choose which logs to stream. After that the client is only
-receiving the frames with actual log messages until it chooses to close the connection. It is important
-for the client to properly terminate the connection as tailing service depends on the close event to 
-release resources used for this service.
+### 2. Recurring jobs
```
a-palchikov, line 38: What kind of jobs are these - do they run as part of the logging application (in the same binary)?

README.md
```diff
@@ -1,87 +1,53 @@
...
+#### Config synchronizer
 
-Implemented query syntax supports filtering on `containers`, `pods` and `file` names:
+Job that syncs forwarders configurations (Gravity "log-forwarders" k8s configMap to Logrange "lr-forwarder" k8s configMap), so that as soon as Gravity forwarders config is updated Logrange have it's configuration updated as well (within couple minutes).
```
a-palchikov, line 41:
```
Job that syncs the updates to Gravity log forwarder configuration to internal format. Frequency can be configured (can it?) and is ~2min by default.
```

README.md
```diff
@@ -1,87 +1,53 @@
...
-```
-If the query is ill-formed or does not contain any sub-filters (i.e. arbitrary search query) - it is used verbatim.
+#### Queries executor
 
-The log forwarder management endpoint handles a PUT request with a JSON-encoded list of external forwarders
-to configure:
-```shell
-$ curl -H "Content-Type: application/json" -X PUT -d '[{"address":"my.example.com:514","protocol":"tcp"},{"address":"your.example.com:514", "protocol":"udp"}]' http://localhost:8083/v1/forwarders
-```
+Job that runs preconfigured queries by schedule. In particular it is used to keep logs database disk usage within limits and perform truncation when needed.
```
a-palchikov, line 45:
```
Job that runs scheduled queries.
In particular, it is used to keep the database size within limits by periodically trimming older entries.
```

README.md
```diff
@@ -1,87 +1,53 @@
...
-
- - [ ] We pass along structured logging information transparently currently, but could do more with it.
- - [ ] Rsyslog is a pretty cranky protocol, we probably want to get rid of it eventually.
- - [ ] Consider replacing the collector with ES/Graylog, for something more indexable/searchable/extensible.
+ 1. Clone repo:<br/>
+ `git clone git@github.com:gravitational/logging-app.git`
+ 2. Enter logging-app directory:<br/>
+ `cd logging-app`
+ 3. Build Gravity package (specify `OPS_URL` and `VERSION` as needed):<br/>
+ `OPS_URL= VERSION=x.x.x make clean tarball`
```
a-palchikov: This does not need to describe each step:

Build instructions:
```shell
$ git clone git@github.com:gravitational/logging-app.git
$ cd logging-app
$ make clean tarball
```
Prefix the `make` command with `OPS_URL=ops-url` to work with a remote package repository and `VERSION=x.y.z` to build a specific version.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+)
+
+func Run(ctx context.Context, cfg *Config, cl api.Client) error {
+	logger := log4g.GetLogger("adapter")
+
+	adaptr, err := NewAdapter(cfg, cl)
+	if err != nil {
+		return fmt.Errorf("failed to create adapter, err=%v", err)
+	}
+	if err := adaptr.Run(ctx); err != nil {
```
a-palchikov: `Run` is a poor name for the method that does not block (as it implies blocking) but instead starts several goroutines. `Start` would be a better name.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+		return fmt.Errorf("failed to run adapter, err=%v", err)
+	}
+
+	<-ctx.Done()
+	_ = adaptr.Close()
+
+	logger.Info("Shutdown.")
+	return err
```
a-palchikov: `err` is implicitly `nil`

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+//===================== adapter =====================
```
a-palchikov: We don't use positional markers - these are just noise.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+func NewAdapter(cfg *Config, cli api.Client) (*Adapter, error) {
+	f := new(Adapter)
+	f.cfg = deepcopy.Copy(cfg).(*Config)
+
+	f.lCli = cli
+	f.logger = log4g.GetLogger("adapter")
+	return f, nil
+}
+
+func (ad *Adapter) Run(ctx context.Context) error {
```
a-palchikov, line 82: This should have a doc comment that should explain that this method is non-blocking and starts a number of goroutines and that the specified context controls their lifespan.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+	ad.runServeAPI(ctx)
+	ad.runSync(ctx)
+	ad.runCronQueries(ctx)
+	return nil
+}
+
+func (ad *Adapter) init() error {
+	var err error
+	ad.kCli, err = newKCli(ad.cfg)
```
a-palchikov: Please use descriptive variable/function names - `newKCli` is a poor choice.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	return err
+}
+
+func (ad *Adapter) Close() error {
+	var err error
+	if !utils.WaitWaitGroup(&ad.waitWg, time.Minute) {
+		err = errors.New("close timeout")
+	}
+	ad.logger.Info("Closed, err=", err)
+	return nil
```
a-palchikov, line 120: Do not ignore the error - leave this up to the caller.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+type (
+	Adapter struct {
+		cfg   *Config
+		wTmpl *forwarder.WorkerConfig
+
+		lCli api.Client
+		kCli *kCli
+
+		waitWg sync.WaitGroup
```
a-palchikov: nit: Usually this is abbreviated as just wg - otherwise it is redundant (i.e. waitWaitGroup)

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+//===================== adapter.api =====================
+
+func (ad *Adapter) runServeAPI(ctx context.Context) {
+	ad.logger.Info("Running serve API on ", ad.cfg.Gravity.ApiListenAddr)
+	ad.waitWg.Add(1)
+	go func() {
+		srv := ad.serveAPI(ctx)
+		select {
+		case <-ctx.Done():
+			sctx, _ := context.WithTimeout(context.Background(),
```
a-palchikov: Do not ignore the `cancel` return - cancel it either after the operation of with defer:
```go
ctx, cancel := context.WithTimeout(context.Background(), ...)
defer cancel()
```

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+				time.Second*time.Duration(5))
```
a-palchikov: `5 * time.Second` is already a time.Duration

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+			if err == nil {
+				_, err = rw.Write(bb)
+			}
+		}
+	}
+
+	if err != nil {
+		ad.logger.Error("log(): Err=", err)
+		rw.WriteHeader(http.StatusInternalServerError)
+		_, _ = rw.Write([]byte("oops, something went wrong!\n"))
```
a-palchikov: We try to avoid cute comments/log messages. I think it's ok to duplicate the http code as a message:
```go
rw.Write([]byte("internal server error\n"))
```

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	return err
+}
+
+func (ad *Adapter) toGrLogEntries(evs []*api.LogEvent) ([]string, error) {
+	gle := make([]string, 0, len(evs))
```
a-palchikov: gle -> entries
le -> logEntry
bb -> bytes

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+func (ad *Adapter) downloadHandler(ctx context.Context,
+	rw http.ResponseWriter, rq *http.Request, p httprouter.Params) error {
+
+	gzWriter := gzip.NewWriter(rw)
+	defer gzWriter.Close()
+	tarWriter := tar.NewWriter(gzWriter)
+	defer tarWriter.Close()
+
+	lql := buildLql("", ad.cfg.Logrange.Partition, dlLimitLines, -dlLimitLines)
```
a-palchikov: This function is pretty hard to follow - consider documenting its behavior and moving parts to auxiliary functions.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+	if err == nil {
+		if buf.Len() > 0 {
+			err = writeTar(fnCntr, buf.Bytes(), tarWriter)
+		}
+	}
+
+	if err != nil {
+		ad.logger.Error("download(): Err=", err)
+		_ = writeTar(fnCntr, []byte("oops, something went wrong!\n"), tarWriter)
```
a-palchikov: I don't think reporting errors as compressed tarballs is good practice - I would stream the logs to the local file system, handle the intermediate errors and then attach the file as a tarball.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	rw.Header().Set("Content-Disposition", "attachment; filename=logs.tar.gz")
+	return nil
+}
+
+func (ad *Adapter) makeHandlerWithCtx(ctx context.Context, handler handlerWithCtx) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+		_ = handler(ctx, w, r, p)
```
a-palchikov: I think the purpose of the wrapper should be simplification of communicating of errors back to client.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+}
+
+//===================== adapter.cron =====================
+
+func (ad *Adapter) runCronQueries(ctx context.Context) {
+	if ad.cfg == nil || ad.cfg.Logrange == nil {
+		ad.logger.Warn("No config or Logrange config, skipping running cron queries")
```
a-palchikov: For me, warning is something to be alert about (i.e. error in configuration or omission of required parameters) - if it's ok that the cron configuration is empty, it should not be logged as warning. On the other hand, if it's validating that the required configuration is available - I'd say it's way too late and should be done up the call chain.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+		return
+	}
+
+	ad.logger.Info("Running ", len(ad.cfg.Logrange.CronQueries), " cron queries")
```
a-palchikov: nit: I'd be more interested in logging the queries themselves rather than their amount.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+func (ad *Adapter) runCronQuery(ctx context.Context, cq cronQuery) {
+	ad.logger.Info("Entering into the loop to run \"", cq.Query, "\" every ", cq.PeriodSec, " seconds.")
+	for {
+		sleepDelay := time.Second * time.Duration(cq.PeriodSec)
+		res, err := ad.lCli.Execute(ctx, api.ExecRequest{Query: cq.Query})
+		if err != nil {
+			ad.logger.Warn("Could not connect to the server to run ", cq)
+			sleepDelay = time.Second
```
a-palchikov, line 168: I would have expected some sort of backoff instead of acceleration in case of errors to be honest.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+		}
+	}
+}
+
+//===================== adapter.sync =====================
+
+func (ad *Adapter) runSync(ctx context.Context) {
+	ad.logger.Info("Running sync every ", ad.cfg.SyncIntervalSec, " seconds...")
+	ticker := time.NewTicker(time.Second *
+		time.Duration(ad.cfg.SyncIntervalSec))
```
a-palchikov, line 187: simpler: `time.NewTicker(ad.cfg.SyncIntervalSec * time.Second)`

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+			ad.sync(ctx)
+		}
+		ad.logger.Warn("Sync stopped.")
+		ad.waitWg.Done()
+	}()
+}
+
+func (ad *Adapter) sync(ctx context.Context) {
+	ad.logger.Debug("sync(): Getting Logrange forwarder config...")
+	lrFwdCfg, err := ad.kCli.getLrFwdCfg()
```
a-palchikov: `getLrPwdCfg` -> `getForwarderConfig`
`getGrFwdCfg` -> `getGravityForwarderConfig`
No need to abbreviate when it hurts readability.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,430 @@
...
+
+func (ad *Adapter) mergeFwdConfigs(lrFwdCfg *forwarder.Config, grFwdCfgs []*grForwarderCfg,
+	wCfgTmpl *forwarder.WorkerConfig) (*forwarder.Config, error) {
+
+	lrFwdCfg = deepcopy.Copy(lrFwdCfg).(*forwarder.Config)
+	lrFwdCfg.Workers = make([]*forwarder.WorkerConfig, 0, len(grFwdCfgs))
+
+	for _, grCfg := range grFwdCfgs {
+		wCfg := deepcopy.Copy(wCfgTmpl).(*forwarder.WorkerConfig)
+		if grCfg.Metadata.Name == "" {
```
a-palchikov: There should a separate validation function to keep this function's responsibility of only merging.
If I were implementing this, I would have `mergeForwarderConfig` either take a single gravity configuration or filter the invalid gravity configurations before calling this function.

cmd/adapter/config.go
a-palchikov: I think it's fine as seconds - all conversions I've seen are superfluous and can be removed.

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil, err
+	}
+	return cfg, nil
+}
+
+//===================== config =====================
+
+func NewDefaultConfig() *Config {
+	return &Config{
+		Gravity: &gravity{
```
a-palchikov: It is not too common to export pure data fields when their type isn't - the callers outside the package will not be able to access any attributes anyways.

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+				ListenAddr: "127.0.0.1:9966",
+			},
+		},
+
+		SyncIntervalSec: 5,
+	}
+}
+
+func (c *Config) Apply(other *Config) {
+	if other == nil {
```
a-palchikov, line 97: Why would you accept pointers in the first place? Just accept by value unless there's a proven performance issue.

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+		c.Logrange.apply(other.Logrange)
+	}
+	if other.SyncIntervalSec != 0 {
+		c.SyncIntervalSec = other.SyncIntervalSec
+	}
+}
+
+func (c *Config) check() error {
+	if c.Gravity == nil {
+		return fmt.Errorf("invalid Gravity=%v, must be non-nil", c.Gravity)
```
a-palchikov: I don't see benefits in outputting `nil` values - the message can be simplified to `Gravity configuration is required`

cmd/adapter/config.go
```diff
@@ -0,0 +1,272 @@
...
+	return nil
+}
+
+func (c *Config) String() string {
+	return utils.ToJsonStr(c)
+}
+
+//===================== gravity =====================
+
+func (g *gravity) apply(other *gravity) {
```
a-palchikov: `apply` seems like a substitute to `merge` (more justified name imho, due to exhibited behavior).

cmd/adapter/main.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		cfg.Apply(config)
+	}
+
+	applyArgsToCfg(c, cfg)
+	return nil
+}
+
+func applyArgsToCfg(c *ucli.Context, cfg *Config) {
+	if sa := c.String(argServerAddr); sa != "" {
```
a-palchikov: sa -> addr
aa -> addr

cmd/adapter/query.go
```diff
@@ -0,0 +1,172 @@
...
+	escaper = strings.NewReplacer("\"", "\\\"")
+)
+
+func parseGrQuery(qs string) (*query, error) {
+	q := &query{}
+	err := qParser.ParseString(qs, q)
+	return q, err
+}
+
+func buildLql(grQuery string, partition string, limit int, offset int) string {
```
a-palchikov: Same concern - `Lql` does not say much to me. Is it an abbreviation?

cmd/adapter/query.go
```diff
@@ -0,0 +1,172 @@
...
+		lql.WriteString(" LIMIT ")
+		lql.WriteString(strconv.Itoa(limit))
+	}
+
+	return lql.String()
+}
+
+func buildOrLql(cnd []*orCondition, files *[]string) string {
+	var orLql bytes.Buffer
+	orLql.WriteString("")
```
a-palchikov: what is this for?

resources/app/log-collector.yaml
```diff
@@ -0,0 +1,148 @@
...
+kind: Service
+metadata:
+  name: log-collector
+  namespace: kube-system
+  labels:
+    app: log-collector
+spec:
+  type: ClusterIP
+  ports:
+    - name: rest
```
a-palchikov: nit: I swear I spent a good minute trying to figure out what this meant. I'd suggest `api` or `web`

resources/app/lr-aggregator.yaml
```diff
@@ -0,0 +1,124 @@
...
+      "HostRegisterTimeoutSec": 0,
+      "PublicApiRpc": {
+        "ListenAddr": "0.0.0.0:9966",
+        "TlsEnabled": false,
+        "Tls2Way": false,
+        "TlsCertFile": "/opt/logrange/config/tls/server.crt",
+        "TlsKeyFile": "/opt/logrange/config/tls/server.key",
+        "TlsCAFile": "/opt/logrange/config/tls/ca.pem"
+      },
+      "JrnlCtrlConfig": {
```
a-palchikov, line 41: I really advise against abbreviations unless they're well known or there's a limit on number of characters in the attribute name.

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied, did some code movements around (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: no reason, probably just copy paste, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: yes, it always returns non-nil res

cmd/adapter/config.go
kbabushkin: mmm, somehow we used to have it in seconds, so leaving it as is, until strong objection...

cmd/adapter/config.go
kbabushkin: we already do this, there is special method on every config "check()" which does this checking, once object is created we don't expect fields to be changed except by apply()/merge() methods which do not set nils

cmd/adapter/config.go
kbabushkin: as I mentioned check() method on configs does this check, also I added some tests, applied (not yet merged to this branch)

resources/app/log-collector.yaml
kbabushkin: for now TlsEnabled=false and those Tls related files are not read, it's just an example how they can be specified here. How exactly they will appear in container (the safe way), this is a good question to figure out, easiest (and not very safe) just mount them...

resources/app/lr-aggregator.yaml
kbabushkin: mmm, this is the question, speaking truly not sure... by default k8s node will not have this /var/lib/logrange folder and when Logrange starts it creates it in /var/lib, so it will be stored on Node (hopefully) and if I mount just /var/lib/logrange it will be just container folder as I understand (i.e. won't survive pod redeploy), might be I'm wrong or there is a way to specify otherwise during mount...

README.md
kbabushkin: applied (not yet merged to this branch)

README.md
kbabushkin: applied (not yet merged to this branch)

README.md
kbabushkin: applied (not yet merged to this branch)

README.md
kbabushkin: applied (not yet merged to this branch)

README.md
kbabushkin: single format, compressed tarball, applied (not yet merged to this branch)

README.md
kbabushkin: added a bit more clarifications, applied (not yet merged to this branch)

README.md
kbabushkin: applied (not yet merged to this branch)

README.md
kbabushkin: applied (not yet merged to this branch)

README.md
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: removed, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: documented, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: renamed, hopefully better names now, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: renamed to wg, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: true, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: done, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: refactored a bit to take the comment into account, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: true, removed that error reporting part, added some comments to the method (not yet merged)...

cmd/adapter/adapter.go
kbabushkin: true, removed not really needed here, ad.cfg or ad.cfg.Logrange can't be nil at this stage (or we already have bigger problems), applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: added printing queries here as well, applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: mmm, doesn't work this way, ad.cfg.SyncIntervalSec is int not time.Duration...

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/adapter.go
kbabushkin: refactored, added filter function, applied (not yet merged to this branch)

cmd/adapter/config.go
kbabushkin: this is intentional for JSON (un-)marshaling, added comment to the struct emphasize this fact...

cmd/adapter/config.go
kbabushkin: removed printing nils, applied (not yet merged to this branch)

cmd/adapter/config.go
kbabushkin: renamed, applied (not yet merged to this branch)

cmd/adapter/main.go
kbabushkin: applied (not yet merged to this branch)

cmd/adapter/query.go
kbabushkin: added more documentation, applied (not yet merged to this branch)

cmd/adapter/query.go
kbabushkin: no reason, some artifact, removed, applied (not yet merged to this branch)

resources/app/log-collector.yaml
kbabushkin: renamed to api, applied (not yet merged to this branch)

resources/app/lr-aggregator.yaml
kbabushkin: took into account, not yet changed

cmd/adapter/config.go
kbabushkin: here no reason, took into account, not yet changed

kbabushkin:
For now just replied/marked the comments here did not yet merged code from `github.com/logrange/logging-app@logrange_with_comments` (which has comments implemented). Will do it tomorrow...

cmd/adapter/adapter.go
kbabushkin: applied (not yet merged to this branch)

resources/app/lr-aggregator.yaml
kbabushkin: ok, I think I got my answer "type: DirectoryOrCreate", will switch to /var/lib/logrange then...

kbabushkin:
Merged branch with applied comments...

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/logrange/logrange/pkg/utils"
+	"sync"
+	"time"
+)
+
+type (
+	// Gravity has certain expectations regarding the interface
+	// its logging application exposes to an end user. Current
+	// logging application implementation is based on Logrange database
+	// which has it's own user interface and which is pretty different from
```
a-palchikov: it's  own -> its own

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	// what is expected by Gravity. The Adapter is intended to be
+	// an actual adapter which sits on top of Logrange interface
+	// and exposes Gravity expected interface.
+	//
+	// Adapter takes the responsibility of making all the needed transformations and
+	// configuration synchronizations in between Gravity and Logrange in order to
```
a-palchikov: ~~in~~ between

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	// meet Gravity logging application requirements.
+	//
+	// Adapter has certain lifecycle and it's caller's responsibility to call
```
a-palchikov: nit: and it's __the__ caller's responsibility to execute Run(ctx) ...

(i.e. I replaced one occurrence of call` with `execute` as an alternative)

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+	if err := ad.init(); err != nil {
+		return trace.Wrap(err)
+	}
+
+	// cancel in case if apiServer fails with error
+	// we need to cancel jobs in terms of this function
+	ctx, cancel := context.WithCancel(ctx)
+
+	// async recurring jobs
+	ad.runSync(ctx)
```
a-palchikov: I'd rename these to `startSync` and `startCronQueries` so it's clear these aren't blocking.

cmd/adapter/adapter.go
```diff
@@ -0,0 +1,196 @@
...
+
+// blocking
+func (ad *Adapter) runApiServer(ctx context.Context) error {
+	ad.logger.Info("Running API server on ", ad.cfg.Gravity.ApiListenAddr)
+	srv := api.NewServer(cfg.Gravity.ApiListenAddr, ad.lrClient, cfg.Logrange.Partition)
+
+	ad.wg.Add(1)
+	go func() {
+		defer ad.wg.Done()
+		select {
```
a-palchikov: you don't need the `select` for a single channel:
```go
<-ctx.Done()
_ = srv.Shutdown()
```

cmd/adapter/adapter.go
a-palchikov: My bad.

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+		server *http.Server
+		// Logrange client query
+		lrClient api.Client
+		// Logrange query partition
+		lrPartition string
+
+		logger *log.Entry
+	}
+
+	// Http request handle but with context support
```
a-palchikov: handle**r**

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+	downloadLinesMax = 500000000
+
+	// Log download limit in bytes per file
+	downloadBytesPerFileLimit = 10 * 1024 * 1024
+
+	// Log download filename prefix
+	downloadFilenamePrfx = "messages"
+)
+
+// Creates new server for the given params,
```
a-palchikov:
```go
// NewServer creates ...
```

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+	return &Server{
+		server:      &http.Server{Addr: listenAddr},
+		lrClient:    lrClient,
+		lrPartition: lrPartition,
+		logger:      log.WithField(trace.Component, "logging-app.api"),
+	}
+}
+
+// Starts serving requests on the configured port, blocking, returns error
+// if underlying htt.Server.Listen() returns err != http.ErrServerClosed
```
a-palchikov: htt**p**.server.

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+	s.server.Handler = router
+	if err := s.server.ListenAndServe(); err != http.ErrServerClosed {
+		return trace.Wrap(err)
+	}
+
+	return nil
+}
+
+// Gracefully shuts down the server,
+// blocks till shutdown, error or timeout happens (10 sec by default)
```
a-palchikov: but the default is not configurable. Instead - I would accept the context from outside.
```go
// Shutdown gracefully shuts down the server.
// It blocks until the server has shut down of context has expired.
func (s *Server) Shutdown(ctx context.Context) error {
```

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+//
+// - 'query':
+//      allowed query terms: "pod", "container", "file", "or", "and"
+//      example: query="pod:p1 and container:c1 and file:f1 or file:f2"
+// - 'limit':
+//      allowed values: int >= 0
+//      example: limit=100
+//
+// In case of error it returns the error (no response write happens) so it's up to
+// caller to handle it properly, e.g. return appropriate http err and code.
```
a-palchikov: appropriate HTTP code

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+//
+func (s *Server) logHandler(ctx context.Context, rw http.ResponseWriter, rq *http.Request, p httprouter.Params) error {
+	var (
```
a-palchikov: nit: `var err error` for single line instead of 3

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+	queryParam := strings.TrimSpace(rq.URL.Query().Get("query"))
+	limitParam := strings.TrimSpace(rq.URL.Query().Get("limit"))
+
+	// validate and set limit
+	limit := defaultTailLinesLimit
+	if limitParam != "" {
+		limit, err = strconv.Atoi(limitParam)
+		if err != nil || limit < 0 {
+			s.logger.Warn("log(): Bad limit=", limitParam, ", using default; err=", err)
+			limit, err = defaultTailLinesLimit, nil
```
a-palchikov: you don't need to reset `err` - it's not read after this

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+	}
+
+	// build Logrange query
+	qr := s.buildQueryRequest(queryParam, "tail", limit, defaultTailLinesOffset)
+	s.logger.Info("log(): Query=", qr.Query)
+
+	// join contexts to handle both server int and transport err
```
a-palchikov: I don't quite follow this comment, what is `server int` - interruption?

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+	jctx, cancel := joincontext.Join(ctx, rq.Context())
+	defer cancel()
+
+	// execute and if executed ok, transform to Gravity format, marshal and write response
+	res := &api.QueryResult{}
+	err = s.lrClient.Query(jctx, qr, res)
+	if err == nil {
```
a-palchikov: You should exit early on error, and do the rest unindented:
```go
if err != nil {
	return trace.Wrap(err)
}
// Handle successful query
```

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+// "/v1/download" api handler, returns compressed tarball stream of logs
+//
+// No query params are supported.
+//
+// In case of error it returns the error so it's up to caller to handle it properly,
+// e.g. return appropriate http err and code.
+//
+// Please note, that if an error occur after a few successful response writes
```
a-palchikov: occur**s**

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+// end user will get file that does not contain all the requested data.
```
a-palchikov: I don't think this is a good contract - I expect either the logs or an error.

cmd/adapter/api/api.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+// Wrapper for http handler, besides calling the actual handler it
+// tries to handle returned errors (if any). In particular,
+// it logs the request, error and writes http error
+func (s *Server) makeHandlerWithCtx(ctx context.Context, handler handlerWithCtx) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+		rw := &responseWriterWithStatus{ResponseWriter: w}
+		err := handler(ctx, rw, r, p)
+		if err != nil {
```
a-palchikov: invert the condition to keep the level of indentation to a minimum:
```go
	rw := &responseWriterWithStatus{ResponseWriter: w}
	err := handler(ctx, rw, r, p)
	if err == nil {
		return
	}
	s.logger.Error("Request=", r, "; err=", err)
	if rw.status == 0 { // write err/status if there were no writes
		trace.WriteError(rw, err)
	}
```

cmd/adapter/config.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/logrange/logrange/pkg/utils"
+	"github.com/logrange/range/pkg/transport"
+	"io/ioutil"
+)
+
+type (
+
+	// Represents Gravity configuration
+	gravity struct {
+		// Address which http server listens
```
a-palchikov: Address **on** which

cmd/adapter/config_test.go
```diff
@@ -0,0 +1,461 @@
...
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/gravitational/logging-app/cmd/adapter/k8s"
+	"github.com/logrange/range/pkg/transport"
+)
+
+func TestLoadCfgFromFile(t *testing.T) {
+	t.Run("test1", func(t *testing.T) {
```
a-palchikov: nit: `test1` is a poor name for a test and does not reveal what's being tested.
Besides I think it's redundant as there's just one test.

cmd/adapter/config_test.go
```diff
@@ -0,0 +1,461 @@
...
+		{name: "test1",
+			fields: fields{
+				Gravity:         newDefaultGravityConfig(),
+				Logrange:        newDefaultLograngeConfig(),
+				SyncIntervalSec: 123,
+			},
+			wantErr: nil,
+		},
+
+		{name: "test2",
```
a-palchikov: For example, instead of the hollow `test2`, I'd write `receives error if gravity config is empty`

cmd/adapter/query/query.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/alecthomas/participle"
+	"github.com/alecthomas/participle/lexer"
+	"strconv"
+	"strings"
+)
+
+type (
+	// Represents parsed 'Gravity log query'
+	query struct {
+		Exp *expression `@@`
```
a-palchikov: This does not conform to struct tag syntax and will issue a warning.
Consider prefixing this with `parser` (See https://github.com/alecthomas/participle/issues/32)

resources/app/log-collector.yaml
r0mant: We need to turn TLS on. All components inside the cluster communicate using mutual TLS.

cmd/adapter/adapter.go
kbabushkin: applied

cmd/adapter/adapter.go
kbabushkin: applied

cmd/adapter/adapter.go
kbabushkin: applied

cmd/adapter/adapter.go
kbabushkin: applied

cmd/adapter/adapter.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: clarified, applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied, added connection close/reset, so that client can detect and handle it...

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/config.go
kbabushkin: applied

cmd/adapter/config_test.go
kbabushkin: applied

cmd/adapter/config_test.go
kbabushkin: applied

cmd/adapter/query/query.go
kbabushkin: applied

resources/app/log-collector.yaml
kbabushkin: @r0mant please let me know how you do it for your existing services, how certs appear in the containers? For now it looks like we need to mount them during container start... (so we need to know from where to take them)... Might be we better have a call to discuss this...

cmd/adapter/api/api.go
```diff
@@ -122,13 +122,10 @@ func (s *Server) Serve(ctx context.Context) error {
...
 
-// Gracefully shuts down the server,
-// blocks till shutdown, error or timeout happens (10 sec by default)
-func (s *Server) Shutdown() error {
-	sctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancel()
-
-	err := s.server.Shutdown(sctx)
+// Shutdown gracefully shuts down the server.
+// It blocks until the server has shut down of context has expired.
```
a-palchikov: of -> or

cmd/adapter/api/api.go
```diff
@@ -197,10 +199,16 @@ func (s *Server) logHandler(ctx context.Context, rw http.ResponseWriter, rq *htt
 // No query params are supported.
 //
 // In case of error it returns the error so it's up to caller to handle it properly,
-// e.g. return appropriate http err and code.
+// e.g. return appropriate HTTP code.
+//
+// Please note, that if an errors occur after a few successful response writes
```
a-palchikov: if error**s** occur
or
if an error occurs

cmd/adapter/config_test.go
```diff
@@ -32,97 +32,71 @@ import (
...
-
-		// check
-		if !reflect.DeepEqual(got, cfg) {
-			t.Errorf("LoadCfgFromFile() = %v, want %v", got, cfg)
-		}
-	})
+	// write
+	cfg := NewDefaultConfig()
+	bb, _ := json.Marshal(cfg)
+	fd, _ := writeTmpJsonFile(bb)
```
a-palchikov: If the write fails (and it's going to be "fun" tracking this down as `os.Remove` call is also unchecked) - will the test output be reasonably clear?

cmd/adapter/k8s/k8s_test.go
```diff
@@ -27,125 +27,95 @@ import (
...
+	grFwdCfgs := []*gravityForwarderCfg{
+		//0
+		{Metadata: struct {
+			Name string `yaml:"name"`
+		}{Name: "name1"},
+			Spec: struct {
+				Address  string `yaml:"address"`
+				Protocol string `yaml:"protocol,omitempty"`
+			}{Address: "127.0.0.1", Protocol: "udp"}},
+		//1
```
a-palchikov: nit: do you need this hint? If you used proper structs, it would look like this:
```go
gravityForwarderConfigs := []*gravityForwarderCfg{
{
	Metadata: Metadata{
		Name: "name1",
	},
	Spec: Spec{
		Address:  "127.0.0.1",
		Protocol: "udp",
	},
},
{
	Metadata: Metadata{
		Name: "name2",
	},
	Spec: Spec{
		Address:  "127.0.0.2",
		Protocol: "",
	},
}}
```
and it would not be so sore on the eye.

a-palchikov:
Since I only have styling nits left - if @r0mant is fine with the state of the implementation, you're free to merge.

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/api/api.go
kbabushkin: applied

cmd/adapter/config_test.go
kbabushkin: added error checking for writeTmpJsonFile()

cmd/adapter/k8s/k8s_test.go
kbabushkin: applied different formatting, tried to make it more readable...

images/adapter/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM debian:stable
```
r0mant: Can you instead use a smaller base image, such as `quay.io/gravitational/debian-tall:stretch`? I don't think container that runs a single binary needs full Debian.

Same for other Docker images (if you have any more, I couldn't see any though).

images/adapter/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apt-get -q -y update && \
+    apt-get install --no-install-recommends -y -q \
+    procps \
+    curl \
```
r0mant: Same as above - I don't think you need all this stuff inside the container.

images/adapter/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    less \
+    dnsutils \
+    netcat \
+    vim
+
+RUN mkdir -p /opt/logrange/gravity
+
+COPY ["build/adapter", "/opt/logrange/gravity/"]
+
+ENTRYPOINT ["/opt/logrange/gravity/adapter", "start", \
```
r0mant: Can you use `dumb-init`? It will make sure signals are passed to your process correctly.

```
ENTRYPOINT ["/usr/bin/dumb-init", "/opt/logrange/gravity/adapter", ...]
```

`dumb-init` is available inside gravitational/debian-tall.

resources/app/lr-collector.yaml
```diff
@@ -0,0 +1,124 @@
...
+            name: lr-collector
+        - name: extdockercontainers
+          hostPath:
+            path: /ext/docker/containers
+        - name: varlibgravitysite
+          hostPath:
+            path: /var/lib/gravity/site
+        - name: varliblogrange
+          hostPath:
+            path: /var/lib/logrange
```
r0mant: Can you make sure to use a path inside `/var/lib/gravity` directory? Otherwise logrange data will not survive upgrades because this directory (/var/lib/logrange) is not mounted from host so it is local to whichever planet is running.

This applies to all resources.

resources/app/lr-forwarder.yaml
```diff
@@ -0,0 +1,99 @@
...
+            path: /var/log
+        - name: varstate
+          hostPath:
+            path: /var/state
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - preference:
+                matchExpressions:
+                  - key: logrange.io/node
```
r0mant, line 97: Who sets this label?

resources/app/log-collector.yaml
```diff
@@ -0,0 +1,146 @@
...
+          "ForwarderConfigMapName": "log-forwarders"
+        }
+      },
+
+      "Logrange": {
+        "Partition": "logrange.pipe=__default__",
+        "ForwarderTmplFile": "/opt/logrange/gravity/config/forward-tmpl.json",
+        "CronQueries": [
+          {
+            "Query": "truncate minsize 4mb maxsize 500mb maxdbsize 10gb",
```
r0mant, line 49: Could you explain this syntax a bit more please? Eg what's 4MB/500MB/10GB? Isn't 10GB a bit too "generous" for the logs? How much data approximately is that?

resources/app/log-collector.yaml
```diff
@@ -0,0 +1,146 @@
...
+            "PeriodSec": 60
```
r0mant, line 50: Isn't that a bit too aggressive schedule for truncating job (just wondering)?

r0mant (commented):
Took a closer look at resources/images now and have a few comments.

resources/app/log-collector.yaml
kbabushkin: **minsize 4mb** - every partition (unique combination of tags) won't be truncated to less than 4 mb, e.g. partition in k8s is a combination of tags: ns (namespace), pod (pod name), cname (container name), cid (container id)...

**maxsize 500mb** - every partition won't be more than 500mb, 'truncate' cronQuery is responsible for this trimming...

**maxdbsize 10gb**  - the whole database won't be more than 10gb (this includes LogEvents, i.e. raw message data + attached tags), again 'truncate' cronQuery is responsible for this trimming...

It's more or less up to you what amount of data you want to store, we can make it less/bigger...
In k8s configuration it can be estimated by taking into account that we store the raw message + 4 tags mentioned above...

resources/app/log-collector.yaml
kbabushkin: Mmm really it depends on how fast the logs are written (to keep them within the bounds)... But in general this doesn't add any obligations in terms of 'global' locks (we don't have this) or anything like that...

kbabushkin: I think 60 sec is fine

resources/app/lr-forwarder.yaml
kbabushkin: It's an artifact from our generic k8s installation where this is done by the installation script. The purpose of this label is to make sure that Aggregator (and Forwarder) is going to be deployed to the same Node as their previous deployments (since they persist state on that node). In Gravity installation we don't apply this label (it is deployment preference not requirement), so probably we should remove it, since I remember we discussed with you that it's uncommon that pod is going to be rescheduled to a different node until manual deletion or node is in a very bad shape or dead...

resources/app/lr-collector.yaml
kbabushkin: will do

images/adapter/Dockerfile
kbabushkin: will do

images/adapter/Dockerfile
kbabushkin: will do

images/adapter/Dockerfile
kbabushkin: will do and yes no other images here

resources/app/lr-collector.yaml
kbabushkin: btw, will all nodes have `/var/lib/gravity` folder, I mean not only masters, right?

resources/app/log-collector.yaml
r0mant: Okay, cool.

resources/app/lr-collector.yaml
r0mant: Yes, all nodes will have it.

kbabushkin: applied

images/adapter/Dockerfile
kbabushkin: applied

images/adapter/Dockerfile
kbabushkin: applied

images/adapter/Dockerfile
kbabushkin: applied

PR #63, opened 2020-01-15
SElinux support

a-palchikov:


resources/app/log-collector.yaml
```diff
@@ -108,10 +108,11 @@ spec:
         app: log-collector
       annotations:
         seccomp.security.alpha.kubernetes.io/pod: docker/default
-
     spec:
       securityContext:
         runAsUser: 0
+        seLinuxOptions:
+          type: gravity_container_logger_t
```
r0mant, line 115: Is this option not going to cause trouble on systems where SELinux is disabled or in permissive mode? Just making sure we don't need to set this conditionally. Cause I know for example that if you have some AppArmor related annotations, the pod won't schedule on nodes where it's unavailable.

r0mant (approved):

resources/app/log-collector.yaml
a-palchikov: That with AppArmor is actually poor experience. I've tested in both modes (with SELinux support on and off but not with SELinux disabled on the host - will validate that) and if docker is not running with `--selinux-enabled`, it neither uses this info nor complains that it cannot do anything with it as Kubelet would be handing it over to the shim unconditionally if I understood correctly.

If this worked only with SELinux enabled on the host - one would have to have multiple resources for each case or handle the SELinux configuration conditionally which I have no idea how to do frankly.

knisbet (approved):

PR #64, opened 2020-03-24
Set requests/limits on logrange components

r0mant:
It's been reported that logrange (specifically, the aggregator) on some clusters consumes a lot of CPU/memory over time (https://github.com/gravitational/gravity/issues/1241).

I've noticed we do not set requests and limits on the logrange components, so this PR sets them. To pick the limits I've looked at the historical data in some of my existing long-running clusters and logrange CPU/memory usage stays stable there (around 1-2% CPU and no more than a 100MB of memory) so I picked something conservative to accommodate potential spikes.

### DIFF
--- a/resources/app/log-collector.yaml
+++ b/resources/app/log-collector.yaml
@@ -132,6 +132,13 @@ spec:
               mountPath: /var/log
             - name: varstate
               mountPath: /var/state
+          resources:
+            requests:
+              memory: 100Mi
+              cpu: 100m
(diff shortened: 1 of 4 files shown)


a-palchikov (approved):

knisbet (approved):
It's good to have this in place, but I'm curious as well about the behaviour if we're getting run away processes. May need to look out for weird behaviour if the process starts running into the limits.

PR #67, opened 2020-07-21
Fix log forwarder resource breaking during upgrades - 5.5.x

eldios:
Fix for https://github.com/gravitational/gravity/issues/1681


images/hook/entrypoint.sh
```diff
@@ -1,10 +1,25 @@
 #!/bin/sh
 set -e
 
-echo "Assuming changeset from the envrionment: $RIG_CHANGESET"
+echo "Assuming changeset from the environment: $RIG_CHANGESET"
 # note that rig does not take explicit changeset ID
 # taking it from the environment variables
-if [ $1 = "update" ]; then
+if [ $1 = "bootstrap" ]; then
```
r0mant, line 7: We don't use rigging for installs. Let's just drop an install script among resources that kubectl's everything, e.g. see https://github.com/gravitational/monitoring-app/blob/version/5.5.x/resources/app.yaml#L23.

eldios: I thought it was better to stay consistent with the flow... any specific reason to not use rigging during the install phase?

r0mant: Well, for one you'd then also need to handle uninstalls properly, otherwise right now the uninstall hook is broken and the phase can't be rolled back / retried if it fails during install. If you want to use rigging, you need to add support for uninstall here as well and make sure the logging-app install phase can be rolled back / retried. I think it might work, although I'm not 100% as we've never really used rigging for installs (it was designed for upgrades), that's why I suggested to keep things simple.

images/hook/entrypoint.sh
```diff
@@ -1,10 +1,25 @@
...
+    echo "Checking: $RIG_CHANGESET"
+    if rig status $RIG_CHANGESET --retry-attempts=1 --retry-period=1s; then exit 0; fi
+
+    echo "Starting bootstrap, changeset: $RIG_CHANGESET"
+    # deleting in case it has been already attempted
+    rig cs delete --force -c cs/$RIG_CHANGESET
+    echo "Creating Log Forwarder ConfigMap"
+    rig upsert -f /var/lib/gravity/resources/init_logforwarder.yaml --debug
```
r0mant: Can you rename the file to just `logforwarders.yaml`?

resources/app.yaml
```diff
@@ -12,15 +12,17 @@ hooks:
...
           metadata:
             name: logging-app-bootstrap
           spec:
             restartPolicy: OnFailure
             containers:
               - name: hook
-                image: quay.io/gravitational/debian-tall:0.0.1
-                command: ["/usr/local/bin/kubectl", "apply", "-f", "/var/lib/gravity/resources/resources.yaml"]
+                image: log-hook:0.0.1
+                args: ["bootstrap"]
```
r0mant, line 25: You need to fix the uninstall hook as well, otherwise it's broken with this change.

r0mant (commented):
lgtm if this works

@a-palchikov @UnderGreen could you folks take a look as well?

images/hook/entrypoint.sh
```diff
@@ -1,10 +1,25 @@
...
-if [ $1 = "update" ]; then
+if [ $1 = "bootstrap" ]; then
+    echo "Checking: $RIG_CHANGESET"
+    if rig status $RIG_CHANGESET --retry-attempts=1 --retry-period=1s; then exit 0; fi
+
+    echo "Starting bootstrap, changeset: $RIG_CHANGESET"
+    # deleting in case it has been already attempted
+    rig cs delete --force -c cs/$RIG_CHANGESET
+    echo "Creating Log Forwarder ConfigMap"
+    rig upsert -f /var/lib/gravity/resources/logforwarder.yaml --debug
```
a-palchikov: If I'm reading this correctly, this will still reset the configuration map on install. What happens if the installation specifies logforwarder configuration via `--config` - in which case configuration will be created before the logging-app install hooks executes?
Do we need empty logforwarder configuration in the first place? IIRC, it will be created on demand when adding/updating the configuration with `gravity resource create`.

eldios: I'm pretty sure that if `rigging` find an existing resource will just replace it so if the configuration specified via `--config` is applied BEFORE the `logging-app` it's gonna be lost.

I don't know if we strictly NEED an empty container but I can definitely investigate.

eldios: confirm that YES, the configmap is needed... 
```
Internal Error
configmaps "log-forwarders" not found, details: &StatusDetails{Name:log-forwarders,Group:,Kind:configmaps,Causes:[],RetryAfterSeconds:0,UID:,}
```

eldios: creating an empty ConfigMap restores the functionality via WEB UI

UnderGreen: Does `--config` invoces the same creating of the ConfigMap? If it is the case then just make `if ! kubectl get configmap name >/dev/null 2>&1` around that rig upsert.

eldios: yeah @a-palchikov was referring to that specific use-case.

I guess that testing if it exists and skipping is definitely the way to go.

Let me add that.

eldios:
> `name` was a placeholder. Replace it with real name of configmap. :)

:facepalm: of course I had the text int he clipboard and copied that instead of my test... thanks @UnderGreen

eldios:
after talking with @r0mant he confirmed that resources provided via `--config` will be created after all other Gravity apps are installed, so I removed the check and I also reverted to just using `kubectl` for simplicity.

a-palchikov (approved):

UnderGreen (approved):

PR #68, opened 2020-07-21
Fixed log-forwarder update issue  - Forward porting for 6.1.x and 7.0.x

eldios:
Fix for https://github.com/gravitational/gravity/issues/1681  - Forward porting for 6.1.x and 7.0.x


r0mant (commented):
This is subject to the same comments as in https://github.com/gravitational/logging-app/pull/67.

eldios:
yep working on both

images/hook/entrypoint.sh
```diff
@@ -5,7 +5,30 @@ set -e
...
 
 echo "--> Assuming changeset from the environment: $RIG_CHANGESET"
-if [ $1 = "update" ]; then
+if [ $1 = "install" ]; then
+    echo "Checking: $RIG_CHANGESET"
+    if rig status $RIG_CHANGESET --retry-attempts=1 --retry-period=1s; then exit 0; fi
+
+    echo "Starting installation, changeset: $RIG_CHANGESET"
+    # deleting in case it has been already attempted
+    rig cs delete --force -c cs/$RIG_CHANGESET
```
a-palchikov: This is going to remove all changes made so far, so the restart of the hook (eg after some resources have already been added/updated) will use the updated resources instead of the original which is not what we want here.

eldios: you shouldn't restart the install hook without first rolling back, right? our steps are never idempotent as far as I know

a-palchikov: The install hook will be restarted automatically (due to the configured policy of `OnFailure`), so the next time the hook runs and provided that it has failed after having already introduced (some) new resources, this will remove the changeset that was previously started and use the new resources as original values if I understand correctly.

eldios: afaik rigging will remove its own changes (thus resetting the configmap as it was created via `--config`) and then restart from there.

I don't see how this could fail... maybe I'm missing something.

a-palchikov: The command `rig cs delete` simply deletes the changset CRD - it does not revert any changes made so far (it's what `rig revert` is for). For install hook I'd even tend towards using `kubectl apply/create` directly as `rig` does not bring any benefits for an install hook especially with restarts after a failure.

PR #72, opened 2020-10-13
Fixed logrange truncating issues

eldios:
Fixed logrange truncating issues.

Updated to version 0.1.46


Makefile
```diff
@@ -4,11 +4,18 @@ NAME := logging-app
 OPS_URL ?= https://opscenter.localhost.localdomain:33009
 OUT ?= $(NAME).tar.gz
 GRAVITY ?= gravity
+LOGRANGE_VERSION := 0.1.46
```
r0mant: Shouldn't it be `v0.1.46`? Or is the image published both with and without the "v" prefix?

go.mod
```diff
@@ -9,11 +9,11 @@ require (
 	github.com/googleapis/gnostic v0.2.0 // indirect
 	github.com/gravitational/logrus v0.10.1-0.20180402202453-dcdb95d728db
 	github.com/gravitational/trace v0.0.0-20190218181455-5d6afe38af2b
-	github.com/jrivets/log4g v0.0.0-20171008071556-0d3a418ba12a
+	github.com/jrivets/log4g v0.0.0-20191016233753-c02c5046dc98
 	github.com/json-iterator/go v1.1.6 // indirect
 	github.com/julienschmidt/httprouter v1.2.0
-	github.com/logrange/logrange v0.1.43
-	github.com/logrange/range v0.0.0-20190621064142-ba348b8545e2
+	github.com/logrange/logrange v0.1.46
```
r0mant, line 15: I see the vendored version here changed, but there are no associated code changes?

resources/app/log-collector.yaml
```diff
@@ -38,7 +38,7 @@ data:
         "ForwarderTmplFile": "/opt/logrange/gravity/config/forward-tmpl.json",
         "CronQueries": [
           {
-            "Query": "truncate minsize 4mb maxsize 500mb maxdbsize 10gb",
+            "Query": "truncate maxdbsize 10gb",
```
r0mant, line 41: Also, what is the reason for removing "minsize" and "maxsize" fields?

eldios: reading the docs I get that `minsize` and `maxsize` are used to filter the logs that are considered for truncating.

So if those are not considered (cause they do not comply to the min-max size rules) they could potentially grow bigger than 10GB, as a whole.

Instead if I moved to use "truncate maxdbsize 10gb" so that no logs will be filtered out, no matter the size of the single partition.

go.mod
eldios: it's the latest fix that Dmitry introduced to workaround apimachinery breaking. That's why the update is there.

Makefile
eldios: fixed.

go.mod
r0mant: Yes, I know, but where are the revendored code changes then?

resources/app/log-collector.yaml
r0mant: Yeah, it makes sense. Let's check with the Logrange team why they put it there in the first place.

Makefile
r0mant: @eldios Going forward, please make sure to test the changes and verify they work before opening a pull request. This issue could have been very easily caught beforehand since there's no such image and it just wouldn't work:

```
➜  ~ docker pull index.docker.io/logrange/logrange:0.1.46
Error response from daemon: manifest for logrange/logrange:0.1.46 not found: manifest unknown: manifest unknown
```

eldios: @r0mant I tested the new logrange version with the hard-coded version and then remembered your suggestion to introduce the "set-image" approach. Thus added that too as a last thing after I opened the PR.

r0mant: Yep, no problem. Shows why it's a good idea to retest after refactoring as well :)

go.mod
a-palchikov: This needs a run of `go mod tidy` / `go mod vendor`

r0mant (approved):

PR #75, opened 2020-12-19
[Feature Request] Expose all log fields and tags contained in log records

eldios:
As per customer request (customer M.) changed code to expose all log fields and tags contained in log records.

### DIFF
--- a/cmd/adapter/api/api.go
+++ b/cmd/adapter/api/api.go
@@ -365,15 +365,22 @@ func toGravityLogEntries(evs []*api.LogEvent) ([]string, error) {
 	return entries, nil
 }
 
+// parseCSVIntoMap takes in a string with a CSV (Comma Separated Value) string
+// with each element in the form `key=value` and translates it into a "dictionary" map
 func parseCSVIntoMap(csv string) (results map[string]string) {
 
 	// iterate over the elements in the form `elem1, elem2, elem3, ...`
 	for _, elem := range strings.Split(csv, ",") {
 
 		// each element should be in the form `key=value`
+		// FIXME this does not play well with keys that contain '='  chars
+		// even if quoted, every other character (whitespace included) is fine
 		elemSplit := strings.Split(elem, "=")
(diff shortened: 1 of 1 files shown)


cmd/adapter/api/api.go
```diff
@@ -357,6 +365,23 @@ func toGravityLogEntries(evs []*api.LogEvent) ([]string, error) {
...
+
+	// iterate over the elements in the form `elem1, elem2, elem3, ...`
+	for _, elem := range strings.Split(csv, ",") {
+
+		// each element should be in the form `key=value`
+		elemSplit := strings.Split(elem, "=")
+		if len(elemSplit) == 2 {
+			results[elemSplit[0]] = elemSplit[1]
+		} else {
+			// one of the element was not in the form `key=value`
```
knisbet, line 385: Is it not valid to have just a key as a tag?

knisbet (approved):

cmd/adapter/api/api.go
```diff
@@ -357,6 +365,23 @@ func toGravityLogEntries(evs []*api.LogEvent) ([]string, error) {
 	return entries, nil
 }
 
+func parseCSVIntoMap(csv string) (results map[string]string) {
+
+	// iterate over the elements in the form `elem1, elem2, elem3, ...`
+	for _, elem := range strings.Split(csv, ",") {
+
+		// each element should be in the form `key=value`
+		elemSplit := strings.Split(elem, "=")
```
a-palchikov, line 378: Can you describe the input format in more detail - maybe as a Godoc for the method?
Can a value be a string with whitespace and potentially include an embedded equals sign if quoted?

cmd/adapter/api/api.go
eldios: Added case where there's a an empty tag with just the key.

Thanks!

cmd/adapter/api/api.go
eldios: explanation comment and FIXME about the '=' char added.

PR #81, opened 2021-03-30
Run containers with normal users

UnderGreen:
* `kubectl rollout status` is added to control the rollout of resources. In prior versions, we just spun up resources without checking errors from them.
* Bump docker source images to newer versions.
* Delete unused volume mounts.
* Configure each component to write on disk into its own directory. They don't interact with each other data, but writing into one `/var/lib/gravity/logrange/lr` causes issues with permissions as `lr-collector` still needs to be run with root to read docker logs.

### Testing done:
- [x] Manual installation with checking all the pods are running
```
log-collector-dddc6d596-rp7d6                1/1     Running     2          43m
logging-app-install-2a65b0-jp5vg             0/1     Completed   0          43m
lr-aggregator-55688457b-j7w5m                1/1     Running     0          43m
lr-collector-9ctdz                           1/1     Running     0          43m
lr-collector-g542k                           1/1     Running     0          43m
lr-collector-psczn                           1/1     Running     0          43m
lr-forwarder-7d848c8545-q9xhp                1/1     Running     0          43m
```

### DIFF
--- a/images/hook/entrypoint.sh
+++ b/images/hook/entrypoint.sh
@@ -14,8 +14,10 @@ if [ $1 = "install" ]; then
     kubectl create -f /var/lib/gravity/resources/app
 
     echo "--> Waiting resources availability"
-    set +e # debug error here
-    kubectl rollout status -f /var/lib/gravity/resources/app
+    for deployment in log-collector lr-forwarder lr-aggregator; do
+	kubectl rollout status deployment $deployment
+    done
+    kubectl rollout status daemonset lr-collector
 elif [ $1 = "update" ]; then
     echo "--> Deleting old deployments"
     for deployment in log-collector lr-forwarder lr-aggregator; do
--- a/images/hook/entrypoint.sh
+++ b/images/hook/entrypoint.sh
@@ -14,6 +14,7 @@ if [ $1 = "install" ]; then
     kubectl create -f /var/lib/gravity/resources/app
 
     echo "--> Waiting resources availability"
+    set +e # debug error here
     kubectl rollout status -f /var/lib/gravity/resources/app
 elif [ $1 = "update" ]; then
     echo "--> Deleting old deployments"
--- a/images/hook/entrypoint.sh
+++ b/images/hook/entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
+set -xe
 
 # note that rig does not take explicit changeset ID
 # taking it from the environment variables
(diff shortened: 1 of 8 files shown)


wadells (commented):
1) Why is change being made (was something breaking before, is this an improvement for compliance?)

2) What is the behavior across upgrades? I don't know logging-app at all, but it seems like the old data will be abandoned in `/var/lib/gravity/logrange/{lr,data}` without any sort of migration or cleanup.

3) Does this need ports?

images/hook/entrypoint.sh
```diff
@@ -1,25 +1,24 @@
...
-set -e
+set -xe
 
 # note that rig does not take explicit changeset ID
 # taking it from the environment variables
 
 echo "--> Assuming changeset from the environment: $RIG_CHANGESET"
 if [ $1 = "install" ]; then
     echo "--> Creating Log Forwarder ConfigMap"
+    # apply is used here because log-forwarder could be created by gravity
```
knisbet, line 10: is the one created by gravity in the resource system? If it does, wouldn't this have the effect of overriding user provided config?

resources/app/log-collector.yaml
```diff
@@ -103,7 +103,7 @@ spec:
     spec:
       priorityClassName: system-cluster-critical
       securityContext:
-        runAsUser: 0
+        runAsUser: -1
```
knisbet, line 106: I thought `-1` was a rigging or gravity think dima added, not sure it works with normal kubectl?

knisbet (commented):
I also share Walts comments about the behaviour through upgrades by changing the path.

UnderGreen:
1. Request from customer M as part of security compliance.

2. We are already sort of abandoning data every upgrade because we don't have persistent storage for the resources. Each time pod spawns on another node it uses hostPath storage from that node. Maybe storage-app in the future fix that problem but for now the application is not persistent and changing the path during the upgrade does not look that bad.

3. Yes it is, forward-port will be provided right after addressing comments here.

resources/app/log-collector.yaml
UnderGreen: I believe gravity tool is doing that job. In this case, it happens in `gravity app unpack` in initContainer for job hook.

images/hook/entrypoint.sh
UnderGreen: I am not sure why this line here. It was here and I didn't change it, just added my thoughts on why it is here. Maybe I am wrong.

resources/app/log-collector.yaml
UnderGreen: Only resources stored as YAML files are subject to automatic translation.  If
a Cluster life cycle hook uses custom resource provisioning, it might need to perform
the conversion manually.

The value of the effective service user ID is stored in the
`GRAVITY_SERVICE_USER` environment variable which is available to each
hook.


From the docs, and source code for sure proves it.

wadells (approved):
> We are already sort of abandoning data every upgrade because we don't have persistent storage for the resources. Each time pod spawns on another node it uses hostPath storage from that node.

Lets at least add docs about this behavior & a release note when this goes out.  It sounds like the patch is broken -- but not significantly more broken than things already were.  Previously we'd sometimes leak/lose some state on upgrade (if the pod was rescheduled on a different host).  Now we always do.

I'm not happy about it, but I also don't see anyone having priority to address it in the holistic sense needed.

Ideally we'd copy the data from the old home to the new one (and vice versa on rollback).

